PR #2, opened 2022-03-24
Setup pytest

bewee:


jb3rndt (changes requested):
I think we should lock the pytest version by adding it to the requirements, shouldnt we?
Rest looks nice

bewee:
That's an interesting question. To my mind, it does not belong into the requirements.txt because it is not required in order to build the application itself: It's only a development dependency. I don't know the python conventions for storing the real application dependencies and dev dependencies however... (Maybe a separate file `dev-requirements.txt`?)

jb3rndt:
Ok I see your point. Yes, a `requirements-dev.txt` is what I saw in some places already, so that is probably the way to go then.

jb3rndt (approved):
Nice thanks 👍

PR #3, opened 2022-03-24
Add basic CI

bewee:


bewee:
I suggest splitting up #4 to move release-related workflows to a separate PR, what do you think?
Also, we considered hosting the app on e.g. heroku and automatically deploying the main branch there

jb3rndt:
Yeah I agree. I'm also not sure if we should split up testing, linting, converage, formatting (is already separate) and building (upload to heroku and so on) into separate workflows to have a more isolated environment for all of them and to directly see what failed?

bewee:
From my experience, build, lint and test are usually put into one workflow together. But I would be fine with separate workflows as well. However, I am also not sure whether this would make it easier for us with this then huge number of pipelines. We need to also keep in mind that each pipeline needs to be set up separately, something that is only required once if we use a combined workflow.

I have put formatting in separate workflows since they use custom gh actions and thus do not require the complete node/python environment setup.

I completely agree with creating a separate workflow for release. I don't completely know yet how this one should look like. I suspect something like building a release and deploying on heroku as soon as a tag is pushed?

jb3rndt:
Ok lets try it how it is now.
For release we have to see what opportunities heroku gives us. Afaik one can connect the github repo to heroku and it is automatically updated on heroku when pushed to a specified branch but im not sure. Building a release would be independent of that then.

jb3rndt (approved):

PR #23, opened 2022-03-28
Feat/10 image display

### LINKED ISSUE #10: Display an Image
- [ ] setup route in backend
- [ ] create HTML page in frontend

jb3rndt:

- Displays an image on the website
- image gets served by the backend
- image folder is `~/.annoto`

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


api/tests/integration/test_integration.py
```diff
@@ -1,5 +1,35 @@
...
+
+
+@pytest.fixture()
+def client(app):
+    return app.test_client()
+
+
+@pytest.fixture()
+def runner(app):
+    return app.test_cli_runner()
```
bewee: These are not used right now, so we may want to consider removing them

bewee (approved):

api/tests/integration/test_integration.py
bewee: Edit: Client is actually used, so consider this comment for runner only

PR #30, opened 2022-03-30
Feat/12 annotation buttons

### LINKED ISSUE #12: Create Annotation
- [ ] Have UI (Buttons) to annotate the image

jdsauer:

- buttons added
- tests for buttons added

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


static/src/components/__tests__/AnnotationButton.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    cy.get("button").should("be.visible");
+  });
+
+  it("is not disabled", () => {
+    mount(AnnotationButton), { probs: { text: String } };
+    cy.get("button").should("not.be.disabled");
+  });
+
+  it("shows specific label", () => {
+    const str = "Label: ";
```
jb3rndt: Maybe it would be useful to use a different `str` here because "Label: " is included in the button by default. So then we could test that `str` is really displayed by the button

static/src/components/__tests__/AnnotationButton.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { mount } from "@cypress/vue";
+import AnnotationButton from "../AnnotationButton.vue";
+
+describe("Button", () => {
+  it("exists", () => {
+    mount(AnnotationButton), { probs: { text: String } };
```
jb3rndt: Maybe I miss something, but what does this test do?

jb3rndt (commented):
Looks nice, just some questions regarding the tests

static/src/components/__tests__/AnnotationButton.spec.ts
jb3rndt: Seems to be convention

n-koenig (approved):

static/src/components/AnnotationButton.vue
```diff
@@ -0,0 +1,35 @@
...
+
+<template>
+  <button @click="log(text)">Label: {{ text }}</button>
+</template>
+
+<style scoped>
+  button{
+    margin: 10px;
+  }
+</style>
```
bewee, line 35: I think there's still some potential here ;) I suppose that's in the other PR?

static/src/components/ImageDisplay.vue
```diff
@@ -1,7 +1,16 @@
...
+ * Displays an image specified by `src`
+ */
+export default defineComponent({
+  props: {
+    /**
+     * Source from where the image should be loaded.
+     */
+    src: { type: String, required: true }
+  },
+})
```
bewee, line 13: Interesting, may I ask what's the benefit of this method over the old one?

bewee (approved):
Looks good!

static/src/components/ImageDisplay.vue
jb3rndt: In theory, either way works the same. Vue has two ways how to define a component with its properties and methods and so on. One is called the Composition API and the other one is the Options API. Both lead to the exact same component but the options api is told to be easier for beginners. Since that sounded good, we used the options api in the AnnotationButton and swapped it out in the ImageDisplay for consistency.
I've not yet dived into that topic very deeply, but here are the links I used for the information on this topic:
https://vuejs.org/guide/introduction.html#api-styles
https://vuejs.org/guide/typescript/composition-api.html
https://vuejs.org/guide/typescript/options-api.html

static/src/components/AnnotationButton.vue
jb3rndt: Yeah definetily, I propose that I fix that in #28 which will require some changes nevertheless ^^

PR #31, opened 2022-03-30
Use black formatter

bewee:
This PR suggests to use black instead of yapf for formatting, since black is more strict concerning the generation of consistent code style

### DIFF
--- a/.github/workflows/format-api.yml
+++ b/.github/workflows/format-api.yml
@@ -6,6 +6,10 @@ on:
       - main
       - dev
 
+defaults:
+  run:
+    working-directory: api
+
 jobs:
   format:
     runs-on: ubuntu-latest
@@ -16,7 +20,6 @@ jobs:
         with:
           repository: ${{ github.repository }}
           token: ${{ github.token }}
-      - run: cd api
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
@@ -34,5 +37,4 @@ jobs:
           git add .
           git commit -m "Automation : Black Formatter" --all | exit 0
       - name: Push changes
-        if: ${{ github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/main' }}
         run: git push
--- a/.github/workflows/format-api.yml
+++ b/.github/workflows/format-api.yml
@@ -18,6 +18,7 @@ jobs:
       - name: checkout repo
         uses: actions/checkout@v2.3.4
         with:
+          ref: ${{ github.event.pull_request.head.ref }}
           repository: ${{ github.repository }}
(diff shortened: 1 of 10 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.formatting.provider": "black"
+}
```
jb3rndt, line 3: Is there a specific reason why we want to have that top level now instead of inside the api folder?

.github/workflows/format-api.yml
```diff
@@ -6,33 +6,36 @@ on:
...
-  yapf:
+  black:
     name: Format api
     runs-on: ubuntu-latest
 
     steps:
       - name: checkout repo
         uses: actions/checkout@v2.3.4
         with:
+          ref: ${{ github.event.pull_request.head.ref }}
```
jb3rndt, line 21: Does that mean it checks out the branch that should be merged? And what was the default before?

.vscode/settings.json
bewee: Actually idk how it ever landed in the api folder in the first place - My vscode always creates the file globally, the other one seems to be unused on my machine. However, if you need it, we can keep both :)

.github/workflows/format-api.yml
bewee: garbage that never worked 😆

.vscode/settings.json
jb3rndt: I moved it manually, but is the one in the subfolder really unused? VsCode has a concept called folder settings so you can override settings for each folder manually and independent of the user settings. So I thought it would fit putting this setting in the api folder, but Im also fine with keeping it at root ^^
(For me there would be no need to keep both)

bewee: Yeah, on my machine it finds the one in the submodule and then copies it into the superfolder^^

I don't really care where we put it either 🤷‍♂️

jb3rndt: Lets leave it how it is now then :D

.github/workflows/format-api.yml
jb3rndt: So we can remove it, right? :D

bewee: Nah, we need it because only this way it works. The old workflow without it would have failed giving some sort of "No HEAD" error. However, the old workflow was actually never executed, because we still checked that we merge into the main branch (what was never the case)^^

jb3rndt: Ah ok now I understand "garbage" referenced xD

jb3rndt (approved):

PR #42, opened 2022-04-05
Feat/19 attentiveness

### LINKED ISSUE #19: proof of attentiveness
- [ ] before saving, display checkbox/dialog for attentiveness
- [ ] saving to JSON

jb3rndt:

- Create Modal and pass data to backend

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


api/mod/src/app.py
```diff
@@ -37,17 +39,49 @@ async def get_image(src: str) -> FileResponse:
 
 
 class Annotation(BaseModel):
```
bewee: To be honest I am still not a huge fan of this. I think having some optional arguments that however have to be always present in some case (here: storing the annotation) is worse than using subclassing. Your approach may make it way harder to find errors, since the presence of src in an annotation stored on the disk can no longer be statically checked by mypy.
Additionally, the Model now contains a property `src` that should never be used (see screenshot below from the API docs). This field will also be available in the auto-generated REST client in the vue app! It suggests that you can put something meaningful in this src field when sending a get request - However it will never be used but overwritten immediately.
image

api/mod/src/app.py
```diff
@@ -37,17 +39,49 @@ async def get_image(src: str) -> FileResponse:
...
+    """
+    An annotation of a file
+
+    Attributes
+        src             name of the annotated file
+        label           label the file [src] should be annotated with
+        hash            hash of the file
+        competency      the competencies the annotator has
+        is_attentive    whether the annotator said that he is attentive
+    """
```
bewee: I'm not sure, but maybe if we put the doc comments on top of each variable, fastapi could be able to use them for its documentation too?

api/mod/src/app.py
```diff
@@ -37,17 +39,49 @@ async def get_image(src: str) -> FileResponse:
...
+
+    def proofs_are_valid(self) -> bool:
+        """Checks whether the proofs of the annotator are all valid"""
+        return self.is_attentive
+
+    def save(self) -> None:
+        """Saves this annotation to the filesystem"""
+        annotation_file = SETTINGS.data_folder.joinpath(f"{self.src}.annotation.json")
+        with open(annotation_file, "w", encoding="utf-8") as file:
+            file.write(json.dumps(self.__dict__, indent=4))
```
bewee: Really great work, so much better!

api/mod/src/app.py
```diff
@@ -60,25 +94,18 @@ class StorableAnnotation(Annotation):
...
+        raise HTTPException(status_code=404, detail="File not found!")
+
+    if not annotation.hash_is_valid():
+        raise HTTPException(
+            status_code=400,
+            detail="Hash values of the provided annotation and the local source do not match!",
+        )
+
+    if not annotation.proofs_are_valid():
+        raise HTTPException(status_code=420, detail="Provided proofs are not valid!")
```
bewee: Please add this response to the decorator documenting this route

api/mod/tests/integration/test_integration.py
```diff
@@ -54,8 +54,11 @@ def test_post_image(client: TestClient) -> None:
     response = client.post(
         "/images/sloth.jpg",
         json={
+            "src": "sloth.jpg",
```
bewee, line 40: Aha! So you even tricked yourselves into using this not-to-be-used attribute :sweat_smile: ^^

api/mod/tests/integration/test_integration.py
```diff
@@ -76,8 +80,11 @@ def test_post_unknown_image_returns_404(client: TestClient) -> None:
     response = client.post(
         "/images/unknown_image.jpg",
         json={
+            "src": "unknown_image.jpg",
             "label": "foo",
-            "hash": "unknown",
+            "hash": "wrong",
```
bewee: Hm? :raised_eyebrow: :thinking:

api/mod/tests/integration/test_integration.py
```diff
@@ -89,9 +96,28 @@ def test_post_image_wrong_hash_returns_400(client: TestClient) -> None:
...
+            "competency": "Dr. Dr. med",
+            "is_attentive": True,
         },
     )
     assert response.status_code == 400
     assert response.headers["content-type"] == "application/json"
+
+
+def test_post_image_with_invalid_proof_return_420(client: TestClient) -> None:
+    """Test POST /images/sloth.jpg raises HttpExeption 420"""
```
bewee:
```suggestion
    """Test POST /images/sloth.jpg raises HttpException 420"""
```

api/mod/tests/integration/test_integration.py
```diff
@@ -89,9 +96,28 @@ def test_post_image_wrong_hash_returns_400(client: TestClient) -> None:
...
             "hash": "wrong",
+            "competency": "Dr. Dr. med",
+            "is_attentive": True,
         },
     )
     assert response.status_code == 400
     assert response.headers["content-type"] == "application/json"
+
+
+def test_post_image_with_invalid_proof_return_420(client: TestClient) -> None:
```
bewee: You sure about that error code? :laughing: 
image

api/mod/tests/unit/test_module.py
```diff
@@ -3,6 +3,76 @@
 """
 
 
-def test_unit() -> None:
```
bewee, line 6: `test_module.py` was more of a placeholder filename tbh. How aboout now renaming it to `test_annotation.py` or similar?

api/mod/tests/unit/test_module.py
```diff
@@ -3,6 +3,76 @@
...
+        mocker.patch("mod.src.settings.SETTINGS.data_folder", Path("mock_folder"))
+        assert Path("mock_folder/test.jpg") == annotation.absolute_src
+
+    def test_file_exists(self, annotation: Annotation) -> None:
+        with Patcher(modules_to_reload=[mod.src.settings, mod.src.app]) as patcher:
+            assert not annotation.file_exists()
+            patcher.fs.create_file(
+                mod.src.settings.SETTINGS.data_folder.joinpath("test.jpg")
+            )
+            assert annotation.file_exists()
```
bewee: I don't really understand the use of this patcher thingy here. What does it do? 
Is this to mitigate having to write to files on the disk? If so, why?

api/mod/tests/unit/test_module.py
```diff
@@ -3,6 +3,76 @@
...
+
+    def test_raises_when_missing_src(self, annotation: Annotation) -> None:
+        """If src is not set, absolute_src should throw an error"""
+        annotation.src = None
+        with pytest.raises(AttributeError):
+            annotation.absolute_src
```
bewee: Yeah, well, you wouldn't have to care about any of this and would have prevented another possible runtime error completely if you had just stayed with the original subclassing variant :/

api/mod/tests/unit/test_module.py
```diff
@@ -3,6 +3,76 @@
...
+    ) -> None:
+        with Patcher(modules_to_reload=[mod.src.settings, mod.src.app]) as patcher:
+            patcher.fs.create_file(
+                mod.src.settings.SETTINGS.data_folder.joinpath("test.jpg"),
+                contents="Dummy Content",
+            )
+            annotation.hash = (
+                "a59ce92485a863931af21370b5082eb8a0c258c1cd74ae068db0ad47aeac1344"
+            )
+            assert annotation.hash_is_valid()
```
bewee: We may also want a test for the negative case. What are your thoughts?

static/src/components/AnnotationButton.vue
```diff
@@ -13,6 +13,8 @@ export default defineComponent({
      */
     label: { type: String, required: true },
     src: { type: String, required: true },
+    competency: { type: String, requred: true },
+    isAttentive: { type: Boolean, required: true },
```
bewee: Hm, a little unfortunate that this actually rather "dumb" button now has to know such really not-at-all-related-to-it things. But I don't really have a better idea for now either

static/src/components/ProofOfAttentiveness.vue
```diff
@@ -0,0 +1,68 @@
...
+        <i-checkbox v-model="checked" class="_text:wrap _margin-y:1/2">
+          <span class="_color:dark _position:relative" style="top: -1px">
+            I am attentive and capable of labeling the data.
+          </span>
+        </i-checkbox>
+        <i-button
+          @click="
+            visible = false;
+            $emit('update:isAttentive', true);
+          "
```
bewee: Maybe consider putting this in a separate function instead of mixing it in here? I'm not sure either :shrug:

bewee (changes requested):
Great work, really love this pretty fancy dialog box! Also, it's great that you cannot insert annotations without actively insuring your attentiveness in the post request! I also really enjoyed reading some well-designed tests here 😊

I added a few comments of varaying scope & importance. I'd especially like to encourage you to think through whether the benefits of using one combined annotation class really outweigh the drawbacks of this decision

api/mod/src/app.py
jb3rndt: It didnt work for me, so Im not sure if that is even possible. Maybe we have to leave it out or look into formatting so it looks better in the docs

api/mod/tests/integration/test_integration.py
jb3rndt: Probably copy-paste error 😅

api/mod/tests/integration/test_integration.py
jb3rndt: Yeah that is what you get for using non-standard error codes xD
grafik
There was not really a fitting error code... We could just go with 400, but what do you think of 428, 412, or 402?

api/mod/tests/unit/test_module.py
jb3rndt: Yes we tried to mock the whole filesystem and replace it with a fake filesystem that lives in-memory. There is not really a need for it but it sounded good 😅
We could otherwise override the data_folder setting with a temporary folder.
Actually I think this would make the tests easier because pyfakefs had problems with mocking pathlib (that is why the patchers are needed) and the changes resulting in that maybe outweigh the benefits of pyfakefs...

jb3rndt:
Ok now I indeed see the consequences caused by removing the subclass.
Lets chat about this on Monday, but I quickly want to hold my thoughts here :)
I think the initial reason we wanted to do it different is because the Annotation class was solely for receiving the data from the request and thus acted like a container. Later it had to be copied to a StorableAnnotation attribute by attribute.
As far as I can see the problems are somewhat caused by the scheme of the api routes (the src is passed in the path and the rest as a body although both belong together). So I think for a nice concept we might need to either change the api route or adapt the rest to the api route scheme. For the second option I had an idea:
Lets treat an Annotation as the annotation metadata (including label, user, attentiveness, ...) that is not directly tied to the data file with its filename but just contains a hash of the data it belongs to. The other player in this is the DataPoint (name suggestions welcome) which represents one data point that can be labeled of the dataset. This datapoint is identified by its filename and can be annotated with an Annotation. This annotation needs to have the correct hash.
(obviously, everything depends on how the whole project develops. If e.g. the competency is not sent by the frontend but only read from the database in the backend, this approach might already get to its limits...)

api/mod/tests/integration/test_integration.py
bewee: 428 sounds reasonable to me

api/mod/tests/unit/test_module.py
bewee: Understand, would have been cool. But concerning the overhead (compared to the ~2 lines required currently in the integration tests) I don't really feel it's all worth it.

jb3rndt (commented):
Approved

bewee (approved):

PR #44, opened 2022-04-06
Auto-generated REST client

bewee:
This suggests using openapi-typescript-codegen to auto-generate a REST client for the Vue application. Use `npm run generate-api` to regenerate it.

---

Right now, the official release does not support the generation of routes that return binary contents. However, there is an open PR to implement this ([986](https://github.com/ferdikoomen/openapi-typescript-codegen/pull/986)) which works like a charm. You can simply install that one on your machine instead:
```sh
cd static
git clone https://github.com/loganbenjamin/openapi-typescript-codegen.git
(cd openapi-typescript-codegen; npm i; npm run build)
npm i openapi-typescript-codegen
npm run generate-api
```
Alternatively, we could fork this fork and publish it on npmjs ourselves for now.


static/src/App.vue
```diff
@@ -28,13 +28,13 @@ export default defineComponent({
       <h1 class="_text-align:center">Annoto</h1>
     </i-layout-header>
     <i-layout-content>
-      <ImageDisplay :src="apiUrl + imageUrl" />
+      <ImageDisplay :src="apiUrl + 'images/' + imageSrc + '?'" />
```
jb3rndt: Maybe it would make sense to use the generated `getUrl` function here, if that is possible? Im not sure if the parameters are available

static/src/components/AnnotationButton.vue
```diff
@@ -29,11 +30,10 @@ export default defineComponent({
...
-        let response = await fetch(`${this.src}?`); // FIXME: Hacky solution to ensure that the browser will not use the previously cached response of a img<src> that would lead to CORS errors
-        let blob = await response.blob();
+        let blob = await API.getImage(this.src);
         let buffer = await blob.arrayBuffer();
         let hash = sha256(buffer);
-        await axios.post(this.src, {
+        await API.saveAnnotation(this.src, {
           label: this.label,
           hash: hash,
         });
```
jb3rndt, line 38: Maybe we can create an Annotation object here?

jb3rndt (commented):
Very nice 👍

static/src/App.vue
bewee: Better now?

static/src/components/AnnotationButton.vue
jb3rndt: What I thought does not seem to work xD

jb3rndt (approved):

PR #45, opened 2022-04-06
Feat/16 usernamepw

jdsauer:

- Show username in web app

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


jb3rndt:
Is there a specific reason why the test files for component tests are `.js` instead of `.ts` now?

jdsauer:
> Is there a specific reason why the test files for component tests are `.js` instead of `.ts` now?

it fixed my path errors and it doesn't changed the outcome of the tests

static/src/components/UserInformationLabel.vue
```diff
@@ -0,0 +1,24 @@
...
+  props: {
+    username: { type: String, required: true },
+  },
+});
+</script>
+
+<template>
+  <div>
+    Current logged in user:
+    <label id="userLabel" for="username">{{ username }}</label>
```
bewee: From my understanding, the `for="username"` is only used to create a relation to an input field. So is this really of any use here?

static/src/components/UserInformationLabel.vue
```diff
@@ -0,0 +1,24 @@
...
+  </div>
+</template>
+
+<style>
+label[for="username"] {
```
bewee: (Same here)

static/src/components/__tests__/UserInformationLabel.spec.js
```diff
@@ -0,0 +1,26 @@
...
+describe("UserInformationLabel", () => {
+  it("exists", () => {
+    mount(UserInformationLabel, { props: { username: "" } });
+  });
+});
+
+// Label renders correctly and is visible
+describe("App", () => {
+  it("renders properly", () => {
+    mount(App, { probs: { username: "" } });
```
bewee:
```suggestion
    mount(UserInformationLabel, { props: { username: "" } });
```
This test seems a little fishy to me, I think you rather wanted to do this?

static/src/components/__tests__/UserInformationLabel.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    cy.get("#userLabel").should("be.visible");
+  });
+});
+
+// Label renders correctly and is visible
+describe("App", () => {
+  it("shows username", () => {
+    mount(App, { probs: { username: "" } });
```
bewee:
```suggestion
    mount(UserInformationLabel, { props: { username: "" } });
```

bewee (commented):
Good work!
I'd suggest renaming the test files back to `.ts` so that we don't loose static type checks. You could however try to change some vscode settings so that this will no longer be falsely marked in the IDE.

bewee (approved):

n-koenig (approved):

PR #46, opened 2022-04-07
Add endpoint to serve data folder

### LINKED ISSUE #43: API Endpoint to GET Annotation JSON

bewee:

- Introduces an endpoint that serves the data folder whenever not in production mode

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


bewee:
Just wanted to make sure that our debug routes do not end up inside the generated API. However, I thereby realized that our decorator does not really remove the route in production mode but only returns an error whenever it is called, thus leaving it inside the production schema.
I wasn't able to figure out how to change the decorator to really remove the route. As much as I love the decorator, I would thus rather replace it back with the simple if statement :/ Probably won't look too bad as soon as we start splitting up the app.py anyway

jb3rndt:
Look nice, I just have one question:

I'm not sure if we should disable the route in prod, or was that part of the user story? Because I don't see the need in debug mode as one can just look into his file system. And leaving out any security issues, it would be helpful to view the files in prod, right? We could think about implementing some access control for such routes but maybe it's not that crucial without it right now. What do you think? :)

bewee:
> Look nice, I just have one question:
> 
> I'm not sure if we should disable the route in prod, or was that part of the user story? Because I don't see the need in debug mode as one can just look into his file system. And leaving out any security issues, it would be helpful to view the files in prod, right? We could think about implementing some access control for such routes but maybe it's not that crucial without it right now. What do you think? :)

Not 100% sure, but I understood it that way: That the route should only be available in dev environments.

Yes, you don't necessarily need it locally. But I actually found it useful when working in the dev docker (with it's not-synced home -dir)
And it's definitely no security concern there, whereas in production...

jdsauer (approved):
very nice 👍

PR #56, opened 2022-04-09
Add toast plugin

jb3rndt:

- Shows toast for 3s after annotation got successfully uploaded

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated
- [x] Maybe also add error toast


static/src/components/Toast.vue
```diff
@@ -0,0 +1,83 @@
...
+    visible: {
+      type: Boolean,
+      required: true,
+    },
+  },
+});
+</script>
+
+<template>
+  <div :class="{ animate: visible, '_position:fixed': true }" id="container">
```
bewee: How exactly does this class syntax thing here work? o.O

static/src/components/Toast.vue
```diff
@@ -0,0 +1,83 @@
...
+  left: 50%;
+  transform: translate(-50%, 0px);
+  z-index: 9999;
+}
+
+div.animate {
+  visibility: visible !important;
+  -webkit-animation: fadein 0.5s, fadeout 0.5s 2.5s;
+  animation: fadein 0.5s, fadeout 0.5s 2.5s;
+}
```
bewee: So this style is somewhat hardcoded to the animation duration of 3s from above, is it?

bewee (approved):
Nice branch name, but unfortunately again missing the issue number ;D Still, really like it!

I found only one small flaw: When clicking on the toast or on the x in its top right corner, the toast re-appears ^^

I think we should consider moving this toasting feature to a more public place sometime in the future, so that for instance errors can be visualized the same way. I imagine something like a public `toast()` function that creates a new toast with a given color & text and potentially overwrites the currently active toast. But for now I find it really sufficient the way it is and would love to see it merged :)

static/src/components/Toast.vue
jb3rndt: Yo, was difficult to parameterize, but for a consistent style I think its ok that this cannot really be tweaked that much

static/src/App.vue
```diff
@@ -55,9 +55,7 @@ export default defineComponent({
         v-model:isAttentive="isAttentive"
         v-model:isTrained="isTrained"
       />
-      <i-label-group block>
-        <UserInformationLabel :username="user" />
-      </i-label-group>
+      <UserInformationLabel :username="user" />
```
bewee, line 70: Is there any particular reason for removing it? I think Jasper mentioned it being somewhat important in another PR🤔

static/src/components/ToastNotification.vue
```diff
@@ -0,0 +1,98 @@
...
+
+  data() {
+    return {
+      isActive: true,
+      typeIcon: new Map([
+        ["info", "ink-info"],
+        ["success", "ink-check"],
+        ["warning", "ink-warning"],
+        ["danger", "ink-danger"],
+      ]),
```
bewee: I'd prefer to name it `typeIcons` (in plural)

static/src/components/ToastNotification.vue
```diff
@@ -0,0 +1,98 @@
...
+});
+</script>
+
+<template>
+  <Transition appear>
+    <div v-show="isActive" id="container" class="_position:fixed">
+      <i-alert :dismissible="dismissible" :color="type" size="md">
+        <template #icon>
+          <i-icon :name="icon" />
+        </template>
```
bewee, line 77: What does this "template in template" syntax achieve?

static/src/plugins/toaster.ts
```diff
@@ -0,0 +1,56 @@
+import { type App, h, render, type Plugin } from "vue";
+import type { ToasterApi } from "@/types";
+import type { ToastProps, ToastOptions } from "@/types";
```
bewee: How does this `@/types` syntax work?

static/src/types.ts
```diff
@@ -0,0 +1,39 @@
+export interface ToastOptions {
```
bewee: Could we somehow move this file into the plugin folder, closer to the other toaster stuff?

static/src/types.ts
```diff
@@ -0,0 +1,39 @@
...
+  type?: "success" | "danger" | "warning" | "info";
+  duration?: number;
+  dismissable?: boolean;
```
bewee: Why are these all optional? Shouldn't they rather be required but provide defaults?

static/src/plugins/toaster.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+function createToaster(app: App, globalOptions: ToastOptions): ToasterApi {
+  return {
+    show(message: string, options: ToastOptions = {}) {
+      const mergedOptions = { ...globalOptions, ...options };
+      const props: ToastProps = { message, ...mergedOptions };
+      const div = document.createElement("div");
+      document.body.appendChild(div);
+      const vNode = h(ToastNotification, props);
```
bewee, line 20: What does this `h` function do?

static/src/plugins/toaster.ts
```diff
@@ -0,0 +1,56 @@
...
+      if (app && app._context) {
+        vNode.appContext = app._context;
+      }
+      render(vNode, div);
+      setTimeout(() => {
+        // The Toast just manages hiding itself but will get removed from the DOM here
+        if (div) {
+          render(null, div);
```
bewee: lol^^

static/src/plugins/toaster.ts
```diff
@@ -0,0 +1,56 @@
...
+          document.body.removeChild(div);
+        }
+      }, (options.duration?.valueOf() || 3000) + 1000);
```
bewee: Why plus 1000? 
Also, I'd really prefer the default value closer to the ToastOptions, but I think I'm slowly getting a feeling for why this is not that simple...

bewee: Could we maybe at least move such magic constants to named constants?

bewee (approved):
Still great, but now even greater :D

Just one last question: Could you give a quick overview of the benefits of copying the toaster functionality into our code versus using the toaster library?

static/package.json
```diff
@@ -44,6 +44,7 @@
     "vue-tsc": "^0.31.4"
   },
   "prettier": {
-    "htmlWhitespaceSensitivity": "ignore"
+    "htmlWhitespaceSensitivity": "ignore",
+    "endOfLine": "auto"
```
bewee: What does it do? Should we add it to the vscode config as well?

static/src/plugins/__tests__/toaster.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+
+    app.use(Toaster);
+    app.mount("body");
+    app.config.globalProperties.$toast.success("test message", {
+      duration: 1,
+    });
+
+    expect(document.body.querySelectorAll("div > div#container").length).equal(
+      1
+    );
```
bewee, line 16:
```suggestion
    expect(document.body.querySelectorAll("div > div#container")).not.to.be.empty;
```
Maybe? Idk

static/package.json
jb3rndt: Default is "lf" so with "auto" it will not format line endings to lf and the linter will thus also not complain about crlf. Initially I wanted to have this setting locally so the linter output is ok, but this will also apply for the ci now. For safety I will remove it again, although all files will be converted to lf when pushing by git.

static/src/plugins/__tests__/toaster.spec.ts
jb3rndt: Yeah the equal formats ugly, so would be nice to have another way but your suggestion would also accept 2 or more matches. I think it would be nice if the test also catches that there are not too many toasts created on a single call

PR #79, opened 2022-04-26
Add login page and switch to task view on login

bewee:
### Changes made by this Pull Request

- Introduce Login page

### Definition of Done

- [x] Tests are written and run successful
- [x] Code is documented
- [x] User Manual is updated
- [x] All requirements of the user story are met

PR #80, opened 2022-04-28
Feat/68 more e2e tests

### LINKED ISSUE #68: Improve vue Integration tests
Add...

- [x] click checkbox and close popup
- [x] annotate without closing popup does not work
- [x] clicking button stores annotation
- [x] next image is shown once image has been annotated

bewee:
Based on #79 

- Adds improved vue integration tests

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


jdsauer (approved):
feels much more clean after changing the tests and deleting duplicate tests

static/cypress/integration/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+    { fixture: "task.json" }
+  ).as("get_task");
+}
+
+export function intercept_next_image() {
+  cy.intercept(
+    "GET",
+    `${Cypress.env("API_URL")}/tasks/ecg-qrs-classification-physiodb/next`,
+    {
+      fixture: "nextImage.txt",
```
jb3rndt, line 39: Should this maybe return the sloth.jpg as we have it here now too?

static/cypress/integration/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  config["intercept_login"] ||= intercept_login;
+  config["intercept_get_task"] ||= intercept_get_task;
+  config["intercept_next_image"] ||= intercept_next_image;
+  config["intercept_get_image"] ||= intercept_get_image;
+  config["intercept_store_annotation"] ||= intercept_store_annotation;
+  config["intercept_login"]();
+  config["intercept_get_task"]();
+  config["intercept_next_image"]();
+  config["intercept_get_image"]();
+  config["intercept_store_annotation"]();
```
jb3rndt: I think we can consider putting all intercepts into a separate intercept function that handles this logic. It would have to be called in every test then too, but it would be clearer that we are mocking the whole api away. What do you think?

Apart from that I think it would be nice to have some kind of default intercept config like 
```js
defaultInterceptConfig = {
  intercept_login: intercept_login,
  ...
}

export function login(interceptConfig: { [key: string]: () => void } = {}) {
  const interceptConfig = { ...defaultInterceptConfig, ...interceptConfig };
  ...
}
```

static/cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
...
+// -- This is a child command --
+// Cypress.Commands.add('drag', { prevSubject: 'element'}, (subject, options) => { ... })
+//
+//
+// -- This is a dual command --
+// Cypress.Commands.add('dismiss', { prevSubject: 'optional'}, (subject, options) => { ... })
+//
+//
+// -- This will overwrite an existing command --
+// Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
```
jb3rndt: Is this file created by accident? I think we already have the typescript version of it

static/cypress/integration/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+    cy
+      .spy(
+        {
+          handle: (req: CyHttpMessages.IncomingHttpRequest) => {
+            req.reply({ statusCode: 204 });
+          },
+        },
+        "handle"
+      )
+      .as("store_annotation_spy")
```
jb3rndt: What does this spy thing do? It seems to be only in this intercept

jb3rndt (commented):
Looks nice, it feels good to have some more frontend tests 👍

static/cypress/support/commands.js
bewee: totally

static/cypress/integration/utils.ts
bewee: Yeah, it's a little dirty to solve it using the login function. On the other hand, we will most likely start every single test with calling the login method anyway... Really unsure about this tbh

PR #81, opened 2022-04-28
Feat/76 user database

### LINKED ISSUE #76: Create Database for users
- [x] decide for a database (format)
- [x] create script for creating new users
- [x] create script for loading test data into the database

jb3rndt:

- Introduces a Database for storing users
- For integration tests, a test database is created and every test is run in a separate transaction that is rolled back after the test
- also includes routes for creating users and requesting a user by email (which is unique)

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


api/Procfile
```diff
@@ -1 +1,2 @@
+release: sh -c "alembic upgrade head && python -m scripts.load_test_data"
```
bewee, line 1: Reminder: Test whether this survives re-deploy

api/alembic.ini
```diff
@@ -0,0 +1,103 @@
...
+# version_path_separator = ;
+# version_path_separator = space
+version_path_separator = os  # Use os.pathsep. Default configuration used for new projects.
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+# Change url here for production
+sqlalchemy.url = sqlite:///$HOME/.annoto/db.db
```
bewee: If this will always be overwritten: Remove it

api/alembic.ini
```diff
@@ -0,0 +1,103 @@
...
+
+[handler_console]
+class = StreamHandler
+args = (sys.stderr,)
+level = NOTSET
+formatter = generic
+
+[formatter_generic]
+format = %(levelname)-5.5s [%(name)s] %(message)s
+datefmt = %H:%M:%S
```
bewee, line 100: No idea what all of this does. I suppose it's all auto-generated?

api/mod/alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
bewee, line 1: Documentation 😎👍

api/mod/alembic/versions/774303782e40_initial_schema.py
```diff
@@ -0,0 +1,41 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "users",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("username", sa.String(), nullable=True),
+        sa.Column("email", sa.String(), nullable=True),
+        sa.Column("hashed_password", sa.String(), nullable=True),
+        sa.Column("salt", sa.String(), nullable=True),
```
bewee: Should they really be nullable?

api/mod/alembic/versions/774303782e40_initial_schema.py
```diff
@@ -0,0 +1,41 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "users",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("username", sa.String(), nullable=True),
```
bewee: Do we really have / want to have a username? For a professional context, I'd rather prefer login via E-Mail and then store the full name (with title etc) for greetings, ...

api/mod/alembic/versions/774303782e40_initial_schema.py
```diff
@@ -0,0 +1,41 @@
...
+        sa.Column("email", sa.String(), nullable=True),
+        sa.Column("hashed_password", sa.String(), nullable=True),
+        sa.Column("salt", sa.String(), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index(op.f("ix_users_email"), "users", ["email"], unique=True)
+    op.create_index(op.f("ix_users_id"), "users", ["id"], unique=False)
```
bewee: unique?

api/mod/src/database/crud.py
```diff
@@ -0,0 +1,3 @@
+"""
+This module holds functions to perform CRUD operations on the database
+"""
```
bewee: Let's remove it for now

api/mod/src/database/database.py
```diff
@@ -0,0 +1,23 @@
...
+This module defines database config and access methods
+"""
+
+from sqlalchemy import create_engine
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import sessionmaker
+
+from mod.src.settings import SETTINGS
+
+Base = declarative_base()
```
bewee, line 11: This variable is never used?

api/mod/src/database/database.py
```diff
@@ -0,0 +1,23 @@
...
+
+engine = create_engine(SETTINGS.database_url, connect_args={"check_same_thread": False})
```
bewee, line 13: Is there any reason for `"check_same_thread": False`?

api/mod/src/database/database.py
```diff
@@ -0,0 +1,23 @@
...
+SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
```
bewee: Why prevent autocommit and autoflush?

api/mod/src/database/database.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def get_db() -> sessionmaker:
```
bewee, line 17: I am still a little unfamiliar with yielding things. Does this create a new connection for every call? o.O

api/mod/src/database/db_models.py
```diff
@@ -0,0 +1,77 @@
+"""
```
bewee, line 1: In the future, we may desire to split this up into one file per model

api/mod/src/database/db_models.py
```diff
@@ -0,0 +1,77 @@
...
+class User(Base):
+    """The basic user database model"""
+
+    __tablename__ = "users"
+
+    id: int = Column(Integer, primary_key=True, index=True)
+    username: str = Column(String, unique=True, index=True)
+    email: str = Column(String, unique=True, index=True)
+    hashed_password: bytes = Column(String)
+    salt: bytes = Column(String)
```
bewee: How comes this is `bytes` here, when it was `string` in the migration?

api/mod/src/database/db_models.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @classmethod
+    def with_password(cls, username: str, email: str, password: str) -> "User":
```
bewee: returns string `"User"`? How does this happen?

api/mod/src/database/db_models.py
```diff
@@ -0,0 +1,77 @@
...
+        return db.query(cls).filter(cls.username == username).first()
+
+    @classmethod
+    def get_by_id(cls, db: Session, user_id: int) -> Optional["User"]:
+        """Get a user by id"""
+        return db.query(cls).filter(cls.id == user_id).first()
+
+    def set_password(self, password: str) -> None:
+        """Set the hashed password"""
+        self.salt = os.urandom(16)
```
bewee, line 59: Is this function cryptographically secure?

api/mod/src/database/db_models.py
```diff
@@ -0,0 +1,77 @@
...
+        self.hashed_password = hashlib.pbkdf2_hmac(
+            "sha512", password.encode(), self.salt, 100000
```
bewee: Magic 100000, consider putting in a constant with a reasonable name

api/mod/src/database/db_models.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+
+    def verify_password(self, password: str) -> bool:
+        """Verify the password"""
+        return hmac.compare_digest(
+            self.hashed_password,
+            hashlib.pbkdf2_hmac("sha512", password.encode(), self.salt, 100000),
```
bewee: Same here

api/mod/src/database/db_models.py
```diff
@@ -0,0 +1,77 @@
...
+        user.set_password(password)
+        return user
+
+    @classmethod
+    def create(cls, db: Session, user: UserCreate) -> "User":
+        """Creates a new user and stores it in the database"""
+        db_user = cls.with_password(user.username, user.email, user.password)
+        db.add(db_user)
+        db.commit()
+        db.refresh(db_user)
```
bewee: Why do we need to refresh it here?

api/mod/src/models/task.py
```diff
@@ -1,4 +1,4 @@
-"""Provides classes for Annotations"""
+"""Provides classes for Tasks"""
```
bewee: Well seen! :+1:

api/mod/src/models/user.py
```diff
@@ -0,0 +1,35 @@
...
+
+class User(UserBase):
+    """An already existing user"""
+
+    id: int = Field(..., description="The id of the user", example=1)
+
+    class Config:
+        """Pydanic configuration"""
+
+        orm_mode = True
```
bewee, line 35: What is this meant to do?

bewee: I find it a little confusing that this guy has no password field. It does make sense in the context of the route where this is used. But I feel a little uncertain about the very absolute name "User" (But right now I have no real counter proposal either)

api/mod/src/routers/login.py
```diff
@@ -14,13 +17,19 @@
...
     status_code=204,
     responses={
+        400: {"description": "No user found with that username"},
         401: {"description": "Failed to validate login"},
     },
     operation_id="login",
 )
 async def login(
     login_data: LoginData,
+    db: Session = Depends(get_db),
```
bewee: Does this `Depends` thing ensure that we always use the same database connection? Or what else is this meant to do?

api/mod/src/routers/login.py
```diff
@@ -14,13 +17,19 @@
     "/",
     status_code=204,
     responses={
+        400: {"description": "No user found with that username"},
```
bewee: I think we should definitely **not**  have a separate error code for this: When there is no such user, we simply say that we couldn't validate the login! I don't know any website that differentiates between "unknown user" and "wrong password". I suppose the reason is that having separate responses would be a security breach that could make it possible to obtain a list of usernames via brute force (although that's just a guess)

api/mod/src/routers/users.py
```diff
@@ -0,0 +1,48 @@
+"""Routes for users"""
+
+from fastapi import APIRouter, Depends, HTTPException
+from sqlalchemy.orm import Session
+
+from mod.src.database import db_models
+from mod.src.database.database import get_db
+from mod.src.models import user
```
bewee: Maybe we could consider doing something like the following to make things easier to understand?
```python3
from mod.src.database.db_models import User as DbUser
from mod.src.models.user import User as UserModel
```

api/mod/src/routers/users.py
```diff
@@ -0,0 +1,48 @@
...
+        raise HTTPException(status_code=404, detail="User not found")
+    return db_user
+
+
+@ROUTER.post(
+    "/",
+    response_model=user.User,
+    responses={
+        400: {"description": "Email or username already registered"},
+    },
```
bewee: Just an idea: There should be some way to provide some sort of "model" for this 400 response that makes it possible to differentiate between "username already registered" and "email already registered". That should then also automatically generate appropriate types in the front-end. I just don't know how :sweat_smile:

api/mod/src/routers/users.py
```diff
@@ -0,0 +1,48 @@
...
+)
+def create_user(user: user.UserCreate, db: Session = Depends(get_db)) -> user.User:
+    """Create a user"""
+    db_user = db_models.User.get_by_email(db, email=user.email)
+    if db_user:
+        raise HTTPException(status_code=400, detail="Email already registered")
+    db_user = db_models.User.get_by_username(db, username=user.username)
+    if db_user:
+        raise HTTPException(status_code=400, detail="Username already registered")
+    return db_models.User.create(db=db, user=user)
```
bewee: Why pass `db` by name here, whereas by reference above?

api/mod/src/settings.py
```diff
@@ -19,6 +19,7 @@ class Settings(BaseSettings):
     """
 
     data_folder: Path = Path.home().joinpath(".annoto")
+    database_url: str = f"sqlite:///{Path.home().joinpath('.annoto').joinpath('db.db')}"
```
bewee: Interesting file extension `.db` :thinking: I mean, it obviously works like this, but maybe we should consider using something more descriptive (and I think also more common?) like `.sqlite3`?

bewee: Same applies for the `conftest.py`

api/mod/tests/conftest.py
```diff
@@ -5,20 +5,31 @@
...
 from mod.src.app import APP
+from mod.src.database.database import get_db
+from mod.src.models.user import UserCreate
 from mod.src.settings import SETTINGS
 
+temp_folder = Path(tempfile.mkdtemp(prefix="annoto"))
+SETTINGS.database_url = f"sqlite:///{temp_folder.joinpath('test_db.db')}"
+
+engine = create_engine(SETTINGS.database_url, connect_args={"check_same_thread": False})
+
```
bewee: From other frameworks, I know that it is possible to create a database in-memory. Is something similar maybe also possible in alembic?

Also: Why create it inside a new temp dir instead of inside the existing `SETTINGS.data_folder` (where also the images etc are stored) ?

bewee: Also: Will this database even be empty again for every single test execution? o.O

api/mod/tests/conftest.py
```diff
@@ -30,3 +41,47 @@ def session() -> Generator:
...
+def apply_migrations() -> Generator:
+    """Apply migrations to test database"""
+    config = Config(str(Path.cwd().joinpath("alembic.ini")))
+    config.set_main_option(
+        "script_location", str(Path.cwd().joinpath("mod").joinpath("alembic"))
+    )
+    config.set_main_option("sqlalchemy.url", SETTINGS.database_url)
+    alembic.command.upgrade(config, "head")
+    yield
+    alembic.command.downgrade(config, "base")
```
bewee: Hm, again similarly to the filesystem stuff: We will do this for **every** single unit test - which is most likely not what we really want

api/mod/tests/conftest.py
bewee: Ok, I see, the database will be wiped by rolling back the transaction. That's probably quicker than what I thought of :thinking:

api/mod/tests/conftest.py
```diff
@@ -30,3 +41,47 @@ def session() -> Generator:
...
+
+
+@pytest.fixture
+def user() -> UserCreate:
+    """Creates a basic Annotation"""
+    return UserCreate(
+        username="AnnotoUser#1337",
+        password="password",
+        email="annoto@team-folivora.com",
+    )
```
bewee: I think this really does not belong into our global conftest

api/mod/tests/conftest.py
```diff
@@ -30,3 +41,47 @@ def session() -> Generator:
...
+    connection = engine.connect()
+    transaction = connection.begin()
+    session = scoped_session(
+        sessionmaker(autocommit=False, autoflush=False, bind=connection)
+    )
+
+    def override_get_db() -> scoped_session:
+        return session
+
+    APP.dependency_overrides[get_db] = override_get_db
```
bewee: Whatever this wild thing does :shrug: :sweat_smile:

api/mod/tests/integration/test_login_router.py
```diff
@@ -3,19 +3,36 @@
...
 
-def test_login(client: TestClient) -> None:
+
+def test_login(client: TestClient, db: Session, user: user.UserCreate) -> None:
     """Test POST /login"""
+    db_models.User.create(db=db, user=user)
     response = client.post(
         "/login/",
-        json={"username": "AnnotoUser#1337", "password": "test1234"},
+        json={"username": "AnnotoUser#1337", "password": "password"},
```
bewee: If you really want to change it, then please to something that exists in your database test entries (as before, the sample data was functional during local development / testing)

api/mod/tests/conftest.py
bewee: It's especially confusing in the tests (e.g. in `test_login_router.py`), where you use the fixture from here with the type annotation from `mod.src.models.user`

api/mod/tests/integration/test_login_router.py
```diff
@@ -3,19 +3,36 @@
...
+        json={"username": "AnnotoUser#1337", "password": "wrong_password"},
+    )
+    assert response.status_code == 400
+
+
+def test_invalid_password_returns_401(
+    client: TestClient, db: Session, user: user.UserCreate
+) -> None:
     """Test POST /login raises HttpException 401"""
+    db_models.User.create(db=db, user=user)
```
bewee: Still don't think that this is a good idea

api/mod/tests/unit/test_user.py
```diff
@@ -0,0 +1,24 @@
...
+import hashlib
+
+from mod.src.database import db_models
+
+
+def test_hash_new_password() -> None:
+    """Test if new hash for password is generated correctly"""
+    password = "password"
+    user: db_models.User = db_models.User.with_password("AnnotoUser", "email", password)
+    test_hash = hashlib.pbkdf2_hmac("sha512", password.encode(), user.salt, 100000)
```
bewee: The magic 100000 is back

api/mod/tests/unit/test_user.py
```diff
@@ -0,0 +1,24 @@
...
+Module for unit tests
+"""
+
+
+import hashlib
+
+from mod.src.database import db_models
+
+
+def test_hash_new_password() -> None:
```
bewee: Also, I'm not completely sure about the value of this test, as it does exactly the same thing the implementation does as well? I'd rather hard-code the hash value, but suppose that's not possible because of the salt?

api/pyproject.toml
```diff
@@ -5,6 +5,7 @@ profile = "black"
 disallow_untyped_defs = true
 allow_redefinition = true
 ignore_missing_imports = true
+exclude = ["alembic"]
```
bewee, line 8: Hm...

api/scripts/load_test_data.py
```diff
@@ -0,0 +1,56 @@
...
+with open(file_path) as f:
+    data = f.read()
+    jsondata = json.loads(data, cls=BytesDecoder)
+
+meta = MetaData()
+meta.reflect(bind=engine)
+for table in meta.sorted_tables:
+    Table = get_class_by_tablename(table.name)
+    if not Table:
+        continue
```
bewee, line 52: Shouldn't this at least log a warning or something?

api/scripts/load_test_data.py
```diff
@@ -0,0 +1,56 @@
...
+
+class BytesDecoder(json.JSONDecoder):
+    def __init__(self, *args, **kwargs):
+        json.JSONDecoder.__init__(self, object_hook=self.object_hook, *args, **kwargs)
+
+    def object_hook(self, dct):
+        if "hashed_password" in dct:
+            dct["hashed_password"] = bytes.fromhex(dct["hashed_password"])
+            dct["salt"] = bytes.fromhex(dct["salt"])
+            return dct
```
bewee: This looks a little weirdly hard-coded to me, but honestly I really have no clue what's going on around this

api/mod/src/routers/users.py
```diff
@@ -0,0 +1,48 @@
+"""Routes for users"""
```
bewee, line 1: From what I understood, `fastapi_restful` should help us not having to write these bloat CRUD functions ourselves

bewee (commented):
First of all: A big  :+1: 
But please keep PRs smaller in the future - This was a nightmare to review :see_no_evil:

api/alembic.ini
jb3rndt: Yep its removed

api/alembic.ini
jb3rndt: Yes its all auto-generated, I have no clue either

api/mod/src/database/database.py
jb3rndt: It's used in each model in `db_models` to serve as a base class

api/mod/src/database/database.py
jb3rndt: There is a comment in the fastapi docs, but in short: It disables a check in sqlite, that prevents different threads from using the same database connection. But fastapi might use multiple threads to handle the same request, so this sqlite feature would break that (I guess). (https://fastapi.tiangolo.com/tutorial/sql-databases/#note)

api/mod/src/database/database.py
jb3rndt: Yes I think it creates a new connection for every request, but I think its common practice to have a new session for each request, isnt it?

api/mod/src/database/db_models.py
jb3rndt: Yep definetily

api/mod/src/database/db_models.py
jb3rndt: We overlooked that I think, it is `LargeBinary` now

api/mod/src/database/db_models.py
jb3rndt: So this is a problem in python, that a classmethod cant return its class (from type hints perspective) because the class is not fully defined yet. So one solution is to wrap the type hint into a String, or use the `from future import annotations` thing. Which one would we prefer? (I just remember that we have the annotation import in a different place already, so we could make it consistent)

api/mod/src/database/db_models.py
jb3rndt: Because the database assigns an id to it and we have to retrieve that id from the database after the insert

api/mod/src/models/task.py
jb3rndt: ;D

api/mod/src/models/user.py
jb3rndt: It enables us to directly convert between the pydantic User model and the sqlalchemy User model. Further information: https://pydantic-docs.helpmanual.io/usage/models/#orm-mode-aka-arbitrary-class-instances

api/mod/src/database/database.py
bewee: It is definitely common to use one transaction per request. But from my (little) experience, you would rather reuse the connection. Just think about a MySQL database or whatever may be used in the future. Is it really desirable to authenticate against the database (which includes logging in, i.e. transmitting db admin username/password, which is rather time consuming) for every single request?

bewee: I think my problem is that I have no clue what a "session" should be in the database world

api/mod/src/models/user.py
jb3rndt: I think the idea was, that we never want to send passwords or even the hashed version to the frontend but want to keep it in the backend or even better only in the database. And this behavior is ensured through design here.
And this is how the fastapi docs recommended it, so this might add some importance 😅

api/mod/src/database/db_models.py
bewee: Yes, let's make consistent. I'd prefer this variant to the future thingy, what do you think?

api/mod/src/routers/login.py
jb3rndt: For most requests we will need access to the database. And this "`Depends`" provides that database connection for the route because the route _depends_ on the database. For each request it will create a new session (as fat as I understood it). I dont know the full power of this mechanism yet, but here are the docs https://fastapi.tiangolo.com/tutorial/dependencies/
So i guess in this case its just a helper to ensure the `get_db` function is invoked everywhere where we need it

api/mod/src/database/db_models.py
bewee: I think this goes in the same direction like my comment on [api/mod/src/database/database.py](https://github.com/team-folivora/annoto/pull/81/files/8ff5429a7306f08e86c58bbe28d5202555a05c0a#diff-6e58a7cde20044bc305187cf0bba622c350cc2e068bb4c3f56828ea80d3aa96c): Would autocommit / autoflush free us from having to do this? This seems to me like something one may accidentially forget (just like closing files, ...) 😬

api/mod/src/routers/login.py
jb3rndt: Yep

api/mod/src/models/user.py
bewee: Nice!

api/mod/src/routers/users.py
jb3rndt: Good idea, do we want `DBUser` or `DbUser`? 😅
I think that is also something we might need to agree on

api/mod/src/routers/users.py
n-koenig: what exactly do you mean by that? I think I understand the idea (using somewhat predefined CRUD operations from fastapi_restful instead of defining all of them ourselves?) but didn't really find any specific information on how to do that.

The closest I found was that: https://fastapi-restful.netlify.app/user-guide/class-resource/

api/mod/src/models/user.py
bewee: Yes, totally agree with you, we should not send passwords to the frontend! (this was never my intention 😅)
For me, it's more of a naming problem, because "User" sounds to me like he would have everything related to users
But I'm also fine leaving it as is

api/mod/tests/conftest.py
jb3rndt: The problem with the datafolder is that it is regenerated (in the session fixture) for every test, currently. So that would also require to rebuild the whole database for each test :/

api/mod/tests/conftest.py
jb3rndt: Isnt the session scope for the whole test run? I thought it would only run once

api/mod/tests/unit/test_user.py
jb3rndt: xD

api/mod/src/routers/users.py
bewee: Hm, good question.
From Ruby on Rails, I know the concept of just defining a class that knows how to construct, store, edit and so on itself, and then the API routes for creating a new instance, getting an instance, editing an instance, listing all instances etc got generated automatically based on this.
I somehow thought fastapi-restful works similarly, but does not really look so, does it?

bewee: Class Resource is most similar to what I meant, but it seems not to ease things as much as I would have expected it to :/

api/mod/src/routers/login.py
bewee: okidoki

api/mod/src/routers/users.py
bewee: idc 😄
Choose something you like and then this will be our convention :)

api/mod/tests/conftest.py
jb3rndt: Ok I see the point with the global conftest, but as you pointed out, it is used in two router tests. We could make a conftest for the integration tests folder, would that be better?
I dont think that solves the import thing you mentioned, but I also do not really understand what you mean there 😅

api/mod/tests/conftest.py
bewee: damn
I'd suggest to leave it as it is, also because I have the feeling that we will no longer need to rebuild our data folder for each and every test as soon as annotations are stored in our db as well - and then we could just remove the test fixture that does so and keep this one, right?

api/mod/tests/conftest.py
bewee: completely true, missed that

api/mod/tests/conftest.py
jb3rndt: xD
It comes back to the `Depends` functionality: the `get_db` function is registered in fastapi as a dependency for the route handlers, and we override this dependency here to use a different function for obtaining the database session

api/mod/tests/integration/test_login_router.py
jb3rndt: Ok yeah we will make that fit again

api/mod/tests/integration/test_login_router.py
jb3rndt: Right, but I propose we keep this test with a different assertion to ensure it fails if the user doesnt exist

api/mod/tests/unit/test_user.py
jb3rndt: Yeah I think we should just drop that test

api/pyproject.toml
jb3rndt: Pylint, black and mypy complained about the autogenerated migration files and we thought as they are autogenerated, they dont necessarily need to be checked. However, I just saw that alembic provides hooks to execute on the migration files, so I can see how far I get with that

api/scripts/load_test_data.py
jb3rndt: Yeah seems like a good idea

api/mod/tests/conftest.py
jb3rndt: import ad `DB`

api/mod/tests/integration/test_login_router.py
jb3rndt: different username

api/mod/tests/unit/test_user.py
jb3rndt: mock urandom

api/mod/src/routers/users.py
jb3rndt: We should just drop fastapi_restful

api/mod/src/routers/users.py
jb3rndt: Should we also import UserCreate as UserCreateModel?

jb3rndt: We agreed on renaming `db_models.User` to `DBUser` and `models.user.User` to `UserResponse` and `models.user.UserCreate` to `CreateUserRequest`

api/mod/src/routers/users.py
n-koenig: Yes I thought so too but I didn't really find any good examples how this Class Resource is actually implemented. And as we are quite familiar with the usual routing style we used so far I also think we can just stick with that.

api/pyproject.toml
jb3rndt: black is now registered as a `post_write_hook`, so all revision files will be correctly formatted now. Sadly I found no way for mypy and pylint to auto-correct the files so I think we should keep the alembic folder ignored

api/mod/src/database/db_models.py
```diff
@@ -0,0 +1,75 @@
...
+        """Verify the password"""
+        return hmac.compare_digest(
+            self.hashed_password,
+            hashlib.pbkdf2_hmac(
+                "sha512", password.encode(), self.salt, self.__hash_iterations__
+            ),
+        )
+
+    def __repr__(self) -> str:
+        return f"<User {self.fullname}>"
```
bewee:
```suggestion
```

api/mod/tests/conftest.py
```diff
@@ -5,20 +5,30 @@
...
 
 from mod.src.app import APP
+from mod.src.database.database import get_db
 from mod.src.settings import SETTINGS
 
+temp_folder = Path(tempfile.mkdtemp(prefix="annoto"))
+SETTINGS.database_url = f"sqlite:///{temp_folder.joinpath('test_db.sqlite3')}"
+
+engine = create_engine(SETTINGS.database_url, connect_args={"check_same_thread": False})
+
```
bewee: Consider moving to `apply_migrations` fixture (and rename the fixture)

api/mod/tests/conftest.py
```diff
@@ -30,3 +40,37 @@ def session() -> Generator:
...
+def apply_migrations() -> Generator:
+    """Apply migrations to test database"""
+    config = Config(str(Path.cwd().joinpath("alembic.ini")))
+    config.set_main_option(
+        "script_location", str(Path.cwd().joinpath("mod").joinpath("alembic"))
+    )
+    config.set_main_option("sqlalchemy.url", SETTINGS.database_url)
+    alembic.command.upgrade(config, "head")
+    yield
+    alembic.command.downgrade(config, "base")
```
bewee: Consider also seeding the database here with our `test_data.json`

api/mod/tests/integration/conftest.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@pytest.fixture
+def user() -> CreateUserRequest:
+    """Creates a basic Annotation"""
+    return CreateUserRequest(
+        fullname="Prof. Dr. Folivora",
+        password="password",
+        email="team@folivora.online",
+    )
```
bewee, line 14: If the database was seeded with `test_data.json`, we would only need this fixture for the users router and could move it there

api/mod/tests/conftest.py
bewee: This could heavily streamline our integration tests (not having to create a test user each time, etc)

bewee (approved):

api/mod/src/database/database.py
jb3rndt: Ill just post some docs snapshots here, also for me to re-read if I need to.
https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session
grafik
grafik

https://stackoverflow.com/questions/44191304/what-does-flushing-the-database-mean-also-with-flash
grafik

jb3rndt: So actually, the `autocommit=False` seems to be the default and we can remove that. Additionally, we might want `autoflush=True` (which would then also be the default), so we can remove that too I guess

api/mod/src/database/database.py
jb3rndt: When using the orm style of sqlalchemy (which we do), the session is our communication point to the database (instead of a connection). In the background, the session also requests a connection from the connection pool as soon as queries are issued. The session additionally keeps track of all orm objects we use within that session and automatically keeps them in sync with the database. Apart from that, the session behaves like a normal connection regarding transactions, committing, flushing and so on. For further information: https://docs.sqlalchemy.org/en/14/orm/session_basics.html

jb3rndt: So to come back to the original question yes, it creates a new session for each request. This session is then the context holder for all the database interactions that happen during this request. And this context should be for that request only and every other request should start with a new clean context. So as far as I understand, this is the way it should be used ^^

api/mod/src/routers/users.py
bewee: https://fastapi.tiangolo.com/advanced/additional-responses/
This is how we should actually handle error response documentation, but since this is an issue for almost all our routes, let's move this to a separate PR!

api/mod/tests/conftest.py
n-koenig: Will save this for later (#86)

api/mod/tests/integration/conftest.py
n-koenig: Will also be done in #86

PR #92, opened 2022-05-04
Python scripts

### LINKED ISSUE #84: Create Python Utility Commands
Similar to package.json fro the frontend, create a list of utility commands and scripts for the backend

Could/Should include:
- [ ] Database Migration (create + apply)
- [ ] Clear Database
- [ ] Dump Database contents to JSON
- [ ] Load JSON contents into database
- [ ] Run tests, linter, mypy, isort, black all at once
- [ ] Reset Database (clear + load dummy content)

n-koenig:

- Adds a list of utility scripts for API management similar to 'package.json'

### Definition of Done

- [x] User Manual is updated
- [x] Code is documented


n-koenig:
Further ideas: 
- [x] Move the 3 database scripts from scripts-folder into one script since they each only have one method anyway

api/mod/fixtures/test_data.json
```diff
@@ -1,7 +1,7 @@
 {
     "alembic_version": [
         {
-            "version_num": "981ce91575bc"
+            "version_num": "be60cc3fc1d3"
```
n-koenig: This might have happened due to deleting+recreating database file+migrations, not sure. Maybe we should exclude the `alembic_version`-table from all loading/dumping operations, since it's not actually "useful data" anyway (as I already did in the `empty_db_tables`-script)?

bewee:
"This branch has conflicts that must be resolved"

bewee:
"This branch has conflicts that must be resolved"

api/manage.py
```diff
@@ -0,0 +1,90 @@
...
+def run_linter():
+    """Runs linter"""
+    os.system("pylint mod")
+
+
+def run_tests():
+    """Runs pytest and provides coverage information when '--cov' is provided"""
+    if len(sys.argv) == 3:
+        if sys.argv[2] == "--cov":
+            os.system("pytest mod --cov")
```
bewee: For the CI, we will additionally need a way to export the xml report

api/mod/fixtures/test_data.json
bewee: Sounds good!

bewee (commented):
We should also adopt the CI and docker files

api/mod/fixtures/test_data.json
jb3rndt: Im not sure because this could be information for alembic to know up to which migration the database state is currently in sync (did that make sense?). Anyway, maybe we should look up the docs what this table is for, to prevent any database created from the test data to be un-migrateable.

api/manage.py
```diff
@@ -0,0 +1,92 @@
...
+
+def clear_db():
+    """Clears data rows in database"""
+    empty_db_tables.empty_db()
+
+
+def reset_db():
+    """Clears database entries and loads test data"""
+    clear_db()
+    load_test_data.load()
```
jb3rndt: I think it would be nice to make one more distinction here (i.e. one more command). One command to just reload the test data (this one here) and another command to recreate the database (and load data), so delete database, apply migrations, load test data.

jb3rndt (commented):
Nice 👍 Im looking forward to using it

api/manage.py
```diff
@@ -0,0 +1,92 @@
...
+    """Runs pytest and provides coverage information when '--cov' is provided"""
+    if len(sys.argv) == 3:
+        if sys.argv[2] == "--cov":
+            os.system("pytest mod --cov")
+        elif sys.argv[2] == "--cov-xml":
+            os.system("pytest mod --cov --cov-report=xml")
+        else:
+            print("Wrong parameters")
+    else:
+        os.system("pytest mod")
```
bewee: Why not simply always execute "pytest" with all the additional arguments provided? This way, we could eg also generate cov reports in other formats etc

api/manage.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def format():
+    """Formats API as specified from GitHub CI"""
+    os.system("black . & isort .")
+
+
+def build():
+    """Builds API as specified from GitHub CI"""
+    os.system("mypy mod & pylint mod & pytest mod --cov --cov-report=xml")
```
bewee:
```suggestion
    os.system("mypy mod & pylint mod & pytest mod --cov")
```
I wouldn't want to generate an xml coverage report whenever I run build
Especially, I think that the xml coverage won't even be .gitignored right now

api/manage.py
```diff
@@ -0,0 +1,92 @@
...
+    "migration-create": create_migration,
+    "migration-apply": apply_migration,
+}
+
+if len(sys.argv) == 1:
+    print_commands()
+elif len(sys.argv) >= 3 and sys.argv[1] != "test":
+    print("Too many arguments provided")
+else:
+    commands[sys.argv[1]]()
```
bewee: What do you think about adding an additional `--help` that lists all possible commands with explanations??

api/scripts/config.py
```diff
@@ -0,0 +1,20 @@
...
+
+from mod.src.settings import SETTINGS
+
+file_path = Path.cwd().joinpath("mod").joinpath("fixtures").joinpath("test_data.json")
+
+if not file_path.exists():
+    print("File does not exist")
+    exit(1)
+
+engine = create_engine(SETTINGS.database_url, connect_args={"check_same_thread": False})
```
bewee: Do we want to use this script on heroku as well? Then we should here too do what #108 suggested

api/scripts/dump_data.py
```diff
@@ -19,11 +14,11 @@ def default(self, z):
...
-with open(outfile_path, "w") as f:
-    f.write(json.dumps(result, cls=BytesEncoder, indent=4))
+def dump():
+    """Dumps all entries from the database into 'test_data.json'"""
+    result = {}
+    for table in meta.sorted_tables:
+        if table.name != "alembic_version":
+            result[table.name] = [dict(row) for row in engine.execute(table.select())]
+    with open(file_path, "w") as f:
+        f.write(json.dumps(result, cls=BytesEncoder, indent=4))
```
bewee, line 24: If we add these few lines, the script also remains executable on its own
```suggestion
        f.write(json.dumps(result, cls=BytesEncoder, indent=4))
      
if __name__ == "__main__":
    dump()
```

api/scripts/empty_db_tables.py
```diff
@@ -0,0 +1,14 @@
...
+from scripts.config import meta, session
+
+
+def empty_db():
+    """Emtpies all rows in all tables except for the alembic version"""
+
+    for table in reversed(meta.sorted_tables):
+        if table.name != "alembic_version":
+            session.execute(table.delete())
+    session.commit()
```
bewee, line 14:
```suggestion
    session.commit()

if __name__ == "__main__":
    empty_db()
```

api/scripts/load_test_data.py
```diff
@@ -32,20 +24,17 @@ def object_hook(self, dct):
...
-        for row in jsondata[table.name]:
-            obj = Table(**row)
-            session.add(obj)
+        if table.name != "alembic_version":
+            Table = get_class_by_tablename(table.name)
+            for row in jsondata[table.name]:
+                obj = Table(**row)
+                session.add(obj)
 
     session.commit()
```
bewee, line 40:
```suggestion
    session.commit()

if __name__ == "__main__":
    load()
```

bewee (approved):
Nice work 👍 ! Looking forward to it being merged so that I can adop the docker container and CI workflows

api/scripts/config.py
jb3rndt: It may be a good idea to always import the engine from `database.py` because that should be configured correctly (also applies to the loading and dumping scripts I think. But that might be for another PR, after this one and #108 are through?

bewee: puh... I mean, this one already touched all the files mentioned anyways, so I'd also be fine just adding it in here. But if you would prefer a separate PR, we could also do this

bewee:
Additional commands we will need for CI & Docker:
- [x] db-create (create database, apply all migrations, fill data)
- [x] run (`python3 -m mod`)
- [x] typecheck (`mypy mod`)
- [x] api-dump (`PYTHON_ENV=production python -m mod.__dumpapi__`)

bewee:
- [x] We should ensure that `python manage.py install` also works when alembic etc are not installed :see_no_evil: 

For this: Maybe only import the required scripts in the functions that really use them?

bewee:
- [x] We may want to consider adding a `python manage.py install-production` that does not install `requirements-dev.txt`

api/scripts/config.py
```diff
@@ -0,0 +1,19 @@
...
+from mod.src.settings import SETTINGS
+
+file_path = Path.cwd().joinpath("mod").joinpath("fixtures").joinpath("test_data.json")
+
+if not file_path.exists():
+    print("File does not exist")
+    exit(1)
+
+SessionLocal = sessionmaker(bind=engine)
+session = SessionLocal()
```
bewee: We can just import this from mod.src.database.database instead of creating it here from scratch

Issue #62, opened 2022-04-12
Use `provide` `inject` for AnnotationButton data


jb3rndt:
provide-inject: https://vuejs.org/guide/typescript/composition-api.html#typing-provide-inject
https://logaretm.com/blog/type-safe-provide-inject/

jb3rndt:
`AsyncComponents` are for loading the components definition asynchronously from the server when needed, which is useful for huge applications. So it does not do what we thought it would do

PR #93, opened 2022-05-04
Feat/78 session

### LINKED ISSUE #78: Session Management
- [ ] correct credentials at login page create new session
- [ ] username is written into annotation from session info
- [ ] close (401) other routes except for login page when no user is logged in

bewee:

- Introduce session management via JWTs

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


bewee:
We based all our work on this tutorial: https://testdriven.io/blog/fastapi-jwt-auth/

api/mod/src/settings.py
```diff
@@ -23,6 +23,9 @@ class Settings(BaseSettings):
         f"sqlite:///{Path.home().joinpath('.annoto').joinpath('db.sqlite3')}"
     )
     debug_routes: bool = True
+    jwt_algorithm: str = "HS256"
```
jdsauer, line 26: maybe we could provide some information about the HS256 in case people wants to know more about the security of the backend and session handling

api/mod/src/auth/auth_bearer.py
```diff
@@ -0,0 +1,41 @@
+from typing import Any, Coroutine, Optional
+
+from fastapi import HTTPException, Request
+from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
+
+from .auth_handler import decodeJWT
```
jb3rndt: An absolute import would be more consistent I think

api/mod/src/auth/auth_bearer.py
```diff
@@ -0,0 +1,41 @@
...
+                raise HTTPException(
+                    status_code=403, detail="Invalid authentication scheme."
+                )
+            if not self.verify_jwt(credentials.credentials):
+                raise HTTPException(
+                    status_code=403, detail="Invalid token or expired token."
+                )
+            return credentials
+        else:
+            raise HTTPException(status_code=403, detail="Invalid authorization code.")
```
jb3rndt: I propose a guard clause would look a bit nicer :P
```python
if not credentials:
    raise HTTPException(status_code=403, detail="Invalid authorization code.")
if not credentials.scheme == "Bearer":
    raise HTTPException(
        status_code=403, detail="Invalid authentication scheme."
    )
if not self.verify_jwt(credentials.credentials):
    raise HTTPException(
        status_code=403, detail="Invalid token or expired token."
    )
return credentials
```

api/mod/src/auth/auth_bearer.py
```diff
@@ -0,0 +1,41 @@
...
+        if credentials:
+            if not credentials.scheme == "Bearer":
+                raise HTTPException(
+                    status_code=403, detail="Invalid authentication scheme."
+                )
+            if not self.verify_jwt(credentials.credentials):
+                raise HTTPException(
+                    status_code=403, detail="Invalid token or expired token."
+                )
+            return credentials
```
jb3rndt: In the Tut it is `credentials.credentials` here. Why do we not want to return the jwt directly?

api/mod/src/auth/auth_bearer.py
```diff
@@ -0,0 +1,41 @@
...
+from .auth_handler import decodeJWT
+
+
+class JWTBearer(HTTPBearer):
+    def __init__(self, auto_error: bool = True):
+        super(JWTBearer, self).__init__(auto_error=auto_error)
+
+    async def __call__(
+        self, request: Request
+    ) -> Optional[HTTPAuthorizationCredentials]:
```
jb3rndt: In the method we check if the credentials exist, so the method cant return None, can it?

api/mod/src/auth/auth_bearer.py
```diff
@@ -0,0 +1,41 @@
...
+    def verify_jwt(self, jwtoken: str) -> bool:
+        isTokenValid: bool = False
+
+        try:
+            payload = decodeJWT(jwtoken)
+        except:
+            payload = None
+        if payload:
+            isTokenValid = True
+        return isTokenValid
```
jb3rndt: We could save some lines here by dropping `isTokenValid` and rather just doing `return payload is not None`

api/mod/src/auth/auth_handler.py
```diff
@@ -0,0 +1,37 @@
+import time
+from typing import Dict, Optional
+
+import jwt
+
+from mod.src.settings import SETTINGS
+
+
+class Payload:
```
jb3rndt: When I initially read `Payload` in the other methods, I thought it is a class from fastapi that contains the whole request payload. What do you think of renaming it to something like `JWTPayload`?

api/mod/src/auth/auth_bearer.py
```diff
@@ -0,0 +1,41 @@
...
+                )
+            return credentials
+        else:
+            raise HTTPException(status_code=403, detail="Invalid authorization code.")
+
+    def verify_jwt(self, jwtoken: str) -> bool:
+        isTokenValid: bool = False
+
+        try:
+            payload = decodeJWT(jwtoken)
```
jb3rndt: `decodeJWT` already catches all exceptions, so in theory we dont need the try except here. Or do we want to keep that for safety?

api/mod/src/routers/images.py
```diff
@@ -78,19 +86,30 @@ async def get_image(
...
     annotation_data: AnnotationData,
     task_id: str = Path(..., example="ecg-qrs-classification-physiodb"),
     src: str = Path(..., example="sloth.jpg"),
+    jwt: Optional[HTTPAuthorizationCredentials] = Depends(JWTBearer()),
+    db: Session = Depends(DB),
 ) -> None:
     """Saves the annotation for the specified image"""
 
+    if not jwt:
+        raise HTTPException(status_code=500)
```
jb3rndt, line 103: If I understand correctly, the error for not being authorized should already be thrown in the `JWTBearer.__call__` method. Is that why this is considered an internal server error here? Will that even ever be triggered?

api/mod/src/routers/images.py
```diff
@@ -78,19 +86,30 @@ async def get_image(
...
+
+    user_id = decodeJWT(jwt.credentials).user_id
```
jb3rndt: should we handle that this can return None?

api/mod/src/settings.py
```diff
@@ -23,6 +23,9 @@ class Settings(BaseSettings):
...
+    jwt_secret: str = "fd232ad1ff70a3a874dda48e857b1ef086e474194ce5d2f8"
```
jb3rndt, line 27: I dont exactly know how to handle this secret, but it should not be exposed in github/git, should it?

api/mod/tests/integration/conftest.py
```diff
@@ -3,12 +3,20 @@
...
 @pytest.fixture
 def user() -> CreateUserRequest:
     """Creates a basic Annotation"""
     return CreateUserRequest(
-        fullname="Prof. Dr. Folivora",
-        password="password",
-        email="team@folivora.online",
+        fullname="Prof. Dr. Perry",
+        password="brrrrrrr",
+        email="perry@folivora.online",
```
jb3rndt: Wyld =D
Will be difficult to remember exactly how many "r" will be needed in the pw when writing tests :P

api/mod/tests/integration/test_images_router.py
```diff
@@ -9,21 +9,27 @@
...
 
 
-def test_get_image(client: TestClient) -> None:
+def test_get_image(client: TestClient, authorization: str) -> None:
     """Test GET /tasks/ecg-qrs-classification-physiodb/sloth.jpg"""
-    response = client.get("/tasks/ecg-qrs-classification-physiodb/sloth.jpg")
+    response = client.get(
+        "/tasks/ecg-qrs-classification-physiodb/sloth.jpg",
+        headers={"Authorization": authorization},
+    )
```
jb3rndt, line 17: So we'll need to add that for every test (that requests a route that needs authorization) I guess. Do you think there is a way to create a fixture like `authorized_client` that is a tweaked `TestClient` that automatically sends requests with Authorization header?

api/mod/tests/integration/test_images_router.py
```diff
@@ -32,9 +38,10 @@ def test_post_image(client: TestClient) -> None:
             "hash": "e922903b4d5431a8f9def3c89ffcb0b18472f3da304f28a2dbef9028b6cd205d",
             "competency": "Prof. Dr. Med",
             "is_attentive": True,
-            "username": "Prof. Dr. Folivora",
+            "fullname": "Prof. Dr. Folivora",
```
jb3rndt: Funny that fastapi throws no error, but the fullname key is not included in the `AnnotationData`, right? Also applies to some other tests

api/mod/tests/integration/test_login_router.py
```diff
@@ -28,13 +24,42 @@ def test_unknown_user_returns_401(client: TestClient) -> None:
...
+    )
+    assert response.status_code == 200
+    token = response.json()["access_token"]
+    response = client.get("/tasks", headers={"Authorization": f"Bearer {token}"})
+    assert response.status_code == 200
+
+
+def test_login_token_expires(client: TestClient, mocker: MockerFixture) -> None:
+    """Test POST /login"""
+    mocker.patch("time.time", lambda: 0)
```
jb3rndt: I think you dont need a lambda here. `return_value=0` works too. What do you prefer? For consistency we should change it in other places as well, if we want to keep lambda

api/mod/tests/integration/test_users_router.py
```diff
@@ -13,8 +13,8 @@ def test_get_user_returns_correct_user(
     client: TestClient, db: Session, user: CreateUserRequest
 ) -> None:
     """Test GET /users/1"""
-    DBUser.create(db=db, user=user)
-    response = client.get("/users/1")
+    user = DBUser.create(db=db, user=user)
```
jb3rndt: I think user creation can be removed since it is done in the `db_engine` fixture now

static/src/components/AnnotationButton.vue
```diff
@@ -43,7 +43,6 @@ export default defineComponent({
         await API.saveAnnotation(this.taskId, this.imageId, {
           label: this.label,
           hash: hash,
-          username: this.username,
```
jb3rndt, line 46: Can we remove the username prop now too?

static/src/components/ImageDisplay.vue
```diff
@@ -16,16 +18,25 @@ export default defineComponent({
...
+    };
+  },
+
+  mounted() {
+    this.src = [
+      computedAsync(async () => {
+        let image = await API.getImage(this.taskId, this.imageId);
+        return image ? URL.createObjectURL(image) : undefined;
+      }, undefined),
+    ];
```
jb3rndt, line 33: Sorry I forgot, but this returns the image so we dont have to use the url in the `` tag which does not work due to authorization right? Nice that you were able to integrate that.
A possible solution for the async thing might be vue asyncComponents. But im fine if we put that in a different pr

jb3rndt (commented):
Nice work 👌
Im still trying to get into the topic, so some comments might be useless

api/mod/src/auth/auth_bearer.py
bewee: I did not figure out how else to get the typing correctly :see_no_evil:

api/mod/src/routers/images.py
bewee: Wow, I'd expect mypy to complain about just these things :thinking:

api/mod/src/settings.py
bewee: Yes and no. I think it's ok for our testing/dev system. We should however definitely overwrite it in the `.env.production` file

api/mod/tests/integration/conftest.py
bewee: I may have been quite tired already at this point, so.... well.... I'd say: Open for change :sweat_smile:

api/mod/tests/integration/test_images_router.py
bewee: Neat idea. Don't know whether / how that's possible, but would definitely be a great idea!

api/mod/tests/integration/test_images_router.py
bewee: Presumably because everything it needs is present, and the remainder will just be ignored :shrug:

api/mod/tests/integration/test_login_router.py
bewee: return_value=0 is perfect

static/src/components/AnnotationButton.vue
bewee: Yeah, but I'd suggest moving this to the other front-end related username/fullname PR - What do you think?

static/src/components/ImageDisplay.vue
bewee: Let's move it to a separate issue, there's already one: #62

api/mod/src/settings.py
bewee: Actually, I don't know anything more about it either... So if you have something to add, feel free to just push it :)

api/mod/src/settings.py
bewee: * or using env variables

api/mod/src/auth/auth_bearer.py
jb3rndt: Kk, me neither

api/mod/src/settings.py
jb3rndt: Yep, should we consider that for staging too?

static/src/components/AnnotationButton.vue
jb3rndt: Ok, I referenced it in #95

api/mod/tests/integration/test_images_router.py
jb3rndt: Ok, but lets put it in another pr

jb3rndt (approved):
The pending conversations are far from being mandatory to solve from my point of view, so feel free to merge :)

api/mod/src/settings.py
bewee: Sounds reasonable

bewee:
Let's do it then!

PR #96, opened 2022-05-05
Heroku

### LINKED ISSUE #94: Fix Heroku
Screenshot from 2022-05-05 09-12-48
422 not good

jb3rndt:
Herokuuuuuu

Closes #94

### DIFF
--- a/api/mod/src/database/database.py
+++ b/api/mod/src/database/database.py
@@ -10,7 +10,7 @@ from mod.src.settings import SETTINGS
 
 Base = declarative_base()
 
-engine = create_engine(SETTINGS.database_url, connect_args={"check_same_thread": False})
+engine = create_engine(SETTINGS.database_url)
 SessionLocal = sessionmaker(bind=engine)
 
 
--- a/api/mod/src/settings.py
+++ b/api/mod/src/settings.py
@@ -34,3 +34,7 @@ class Settings(BaseSettings):
 
 
 SETTINGS = Settings()
+if SETTINGS.database_url.startswith("postgres://"):
+    SETTINGS.database_url = SETTINGS.database_url.replace(
+        "postgres://", "postgresql://"
+    )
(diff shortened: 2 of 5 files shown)


api/mod/src/database/database.py
```diff
@@ -10,7 +10,7 @@
 
 Base = declarative_base()
 
-engine = create_engine(SETTINGS.database_url, connect_args={"check_same_thread": False})
```
jb3rndt, line 13: This option is incompatible with postgres. I tested around a bit and it doesnt seem to break anything locally. We might need to keep an eye on that

jb3rndt (commented):
On Heroku, we will use a postgres instance. SQLite does not really work on heroku

api/mod/src/settings.py
```diff
@@ -34,3 +34,7 @@ class Config:
 
 
 SETTINGS = Settings()
+if SETTINGS.database_url.startswith("postgres://"):
+    SETTINGS.database_url = SETTINGS.database_url.replace(
```
bewee, line 38: Why?

bewee (approved):
Just two things:
- Changes in the load_test_data script, but we do not use it in heroku afaics? 🤔 
- We could consider add a setting for that thread stuff which we set to false in production and keep it true for dev. But since you say everything works fine in dev anyway, I'd leave at it as it is and keep it in the back of our minds 👍

api/mod/src/settings.py
jb3rndt: Heroku automatically sets a DATBASE_URL config variable that holds the connection string to the database. Sadly, it is not editable. But sqlalchemy will only recognise the database type when the string starts with `postgresql`, but Heroku writes only "postgres" so this needs to be replaced

jb3rndt:
Yeah the changes to the load script are just for refactoring, which I recognized through doing this. I think its nothing very special so we can just keep it in here?

api/mod/src/settings.py
bewee: okidoki

PR #100, opened 2022-05-05
Prevent data loss when multiple users annotate the same image

bewee:


api/mod/src/models/annotation.py
```diff
@@ -91,6 +91,14 @@ def save(self) -> None:
         if not self.hash_is_valid():
             raise HashMismatch()
         annotation_file = SETTINGS.data_folder.joinpath(f"{self.src}.annotation.json")
+
+        trial = 2
+        while annotation_file.exists():
```
n-koenig, line 96: Just one short question: Could we also try to "read" the current number from the filename (if it exists) and then instantly set the correct one instead of the while loop?

bewee: I don't understand what you mean tbh...

api/mod/src/models/annotation.py
```diff
@@ -91,6 +91,14 @@ def save(self) -> None:
...
+            annotation_file = SETTINGS.data_folder.joinpath(
+                f"{self.src}.annotation.{trial}.json"
+            )
+            trial += 1
+
```
jdsauer, line 101: so you just increment the trial number and save it into the filename, neat and good solution!

jdsauer (approved):
smart and good solution to prevent the overwrite! get my approve 👍

PR #105, opened 2022-05-09
ManagedTestClient with authorization capabilities

### LINKED ISSUE #98: Look into creating an authenticated client for tests
https://github.com/team-folivora/annoto/pull/93#discussion_r865692550

> Do you think there is a way to create a fixture like authorized_client that is a tweaked TestClient that automatically sends requests with Authorization header?

jb3rndt:

- Introduces new standard TestClient

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


api/mod/tests/integration/test_login_router.py
```diff
@@ -54,12 +55,12 @@ def test_login_token_expires(client: TestClient, mocker: MockerFixture) -> None:
     )
     assert response.status_code == 200
     token = response.json()["access_token"]
-    mocker.patch("time.time", returl_value=1e9)
+    mocker.patch("time.time", return_value=1e9)
```
jdsauer, line 58: strange return value, but I guess it is necessary :)

jdsauer (approved):
Nice work, now it should be much easier to write tests which needs auth. I'm still a bit confused about the mocker return codes, but they seem to be that way. Good job!

PR #107, opened 2022-05-09
Change Component Library and write component tests

### LINKED ISSUE #73: Fix Frontend Tests
Frontend-tests currently don't work properly (cypress doesn't properly recognise components from inkline)

Possible fixes:
- look if we find another testing framework for component tests or just drop them (risk of big and slow e2e tests)
- try out whether component tests work without any component library / with another one (eg bootstrap, ...)

jb3rndt:

- Replace Inkline with Element Plus
- More and working component tests

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated

### ToDo
- [ ] Change font
- [x] actually write component tests


static/cypress/integration/utils.ts
```diff
@@ -66,6 +66,14 @@ export function intercept_get_image() {
...
 }
 
+export function intercept_get_image_with_failure() {
+  cy.intercept(
+    "GET",
+    `${Cypress.env("API_URL")}/tasks/ecg-qrs-classification-physiodb/sloth.jpg`,
+    { statusCode: 404 }
+  ).as("get_image");
+}
+
```
bewee, line 76: I don't see how this belongs here? 🤔

jb3rndt: Ah good question: How do we want to organize utility functions? Because they are needed for the component tests and integration tests. We could put the component tests in the cypress folder too, or move the utils out of the integration folder

jb3rndt:
We decided to abandon that (for now) because it causes some issues with the e2e tests and the new component tests do not add much value now and are redundant to the e2e tests.

PR #108, opened 2022-05-10
Fix database connection settings

### LINKED ISSUE #106: BruhSql
grafik

jb3rndt:

### DIFF
--- a/api/mod/src/settings.py
+++ b/api/mod/src/settings.py
@@ -26,7 +26,7 @@ class Settings(BaseSettings):
     jwt_algorithm: str = "HS256"
     jwt_secret: str = "fd232ad1ff70a3a874dda48e857b1ef086e474194ce5d2f8"
     jwt_expiry: int = 60 * 60 * 24  # one day
-    engine_connect_args: dict = {"check_same_thread": False}
+    engine_connect_args: dict = {}
 
     class Config:
         """Reads the dotenv file"""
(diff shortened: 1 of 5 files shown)


api/mod/src/settings.py
```diff
@@ -35,6 +36,7 @@ class Config:
 
 SETTINGS = Settings()
 if SETTINGS.database_url.startswith("postgres://"):
+    SETTINGS.engine_connect_args = {}
```
bewee: I'm somewhat unsure whether that's the best way to do it... I see how it should always be `"check_same_thread": True` when using a postgres database. On the other hand, we could solve this without writing this additional line of code by specifying `engine_connect_args = {"check_same_thread": False}` in the dotenv file.

Besides that: This could overwrite some other part of the `engine_connect_args` when using the dotenv file, right? Then this solution seems a little radical to me - we should only overwrite the `"check_same_thread"` property and leave the rest as it is

bewee (approved):

api/mod/src/settings.py
jb3rndt: Oh I didnt know putting `engine_connect_args = {"check_same_thread": False}` in the dotenv is possible, gonna do that then ^^

jb3rndt: Im not sure but I think postgres already threw an error not because the value is True or False, but because the setting is not valid, so the dict must not contain the key I think

bewee: Okidoki
We could also simply add the `engine_connect_args` option to our `.env.dev` file and leave it unset when in staging?

bewee: So what I mean: Instead of wiping the engine_connect_args in staging + production, only set them in `.env.dev`

jb3rndt: Ah yes

bewee:
Feel free to merge it

PR #110, opened 2022-05-10
Update docker container to use python management

bewee:


jb3rndt (approved):
No clue what it does :D

PR #114, opened 2022-05-11
Feat/72 timestamp

### LINKED ISSUE #72: Timestamps for Annotations
As an annotation provider, I want to be able to audit the annotations.

## Acceptance Criteria
- [ ] each annotation has a timestamp

n-koenig:

- Introduces timestamp-blockchain

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


api/mod/src/models/annotation.py
```diff
@@ -50,14 +50,22 @@ class Annotation(AnnotationData):
         description="The full name of the annotator",
         example="Prof. Dr. Folivora",
     )
+    timestamp: int = Field(
+        ...,
+        description="The current timestamp when the annotation is saved",
+        example=420,
```
bewee: Maybe we should use something more realistic for the example?
(just copy the current timestamp in here :shrug: )

bewee (approved):

api/mod/src/models/annotation.py
bewee:
```suggestion
        example=1652263830,
```

PR #115, opened 2022-05-11
Fix/pylint version

n-koenig:

- Sets the correct requirement so that pylint actually works

**To upgrade the package, run: `pip install --upgrade pylint` in your venv or docker!**


n-koenig:
I made this a new pull request and as soon as #114 is merged this will only be a single file change

bewee (approved):
Greatest piece of work I've ever seen!

PR #119, opened 2022-05-12
Feat/117 annotations list

### LINKED ISSUE #117: Page that displays available annotation tasks

jb3rndt:

- List available tasks
- Open annotation view for one selected task

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


api/mod/src/models/task.py
```diff
@@ -7,7 +7,7 @@
 
 class Task(BaseModel):
     """
-    A labelling task
+    A labeling task
```
bewee, line 10: I think labelling is gramatically correct, it's the gerund. Both forms should be valid here.

https://www.thefreedictionary.com/labeling

bewee:
But we should definitely look into the router thingy in another PR :sweat_smile:

static/cypress/integration/tasks_overview.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    setup_intercepts();
+    login().get("#userLabel").contains("Prof. Dr. Folivora");
+  });
+
+  it("shows cards for tasks", () => {
+    setup_intercepts();
+    login().get(".card#task-ecg-qrs-classification-physiodb");
+  });
+
+  it("Informs the user when there are no tasks", () => {
```
jb3rndt: why capitalized?

api/mod/tests/integration/test_tasks_router.py
```diff
@@ -12,7 +12,24 @@ def test_get_tasks(client: ManagedTestClient) -> None:
     response = client.get("/tasks")
     assert response.status_code == 200
     assert response.headers["content-type"] == "application/json"
-    assert response.json() == ["ecg-qrs-classification-physiodb"]
+    assert response.json() == [
```
n-koenig, line 15: Suggestions from Meeting: Add more detailed method descriptions and write out pylint code, maybe also test if there is a correct response if no tasks are available

api/mod/tests/integration/test_tasks_router.py
```diff
@@ -12,7 +12,24 @@ def test_get_tasks(client: ManagedTestClient) -> None:
     response = client.get("/tasks")
     assert response.status_code == 200
```
n-koenig, line 14: might be unnecessary if `response.json()` from line 15 already tests this

api/mod/src/routers/tasks.py
```diff
@@ -17,21 +17,30 @@
...
-        }
-    },
+    response_model=List[Task],
     operation_id="get_tasks",
     dependencies=[Depends(JWTBearer())],
 )
-async def get_tasks() -> List[str]:
-    """Get a list of the IDs of all available labelling tasks"""
+async def get_tasks() -> List[Task]:
+    """Get a list of the IDs of all available labeling tasks"""
```
n-koenig:
```suggestion
    """Get a list of all available labeling tasks"""
```

static/cypress/fixtures/tasks.json
```diff
@@ -0,0 +1,13 @@
...
+    "description": "Labeling the QRS complex of Physionet ECGs",
+    "labels": [
+      "Atrial fibrillation",
+      "Normal sinus rhythm",
+      "Ventricular tachycardia",
+      "Noise",
+      "Other"
+    ]
+  }
+]
```
n-koenig, line 13: is this missing the second task or am I missing something here?

n-koenig: Also, we now have a `task.json` and a `tasks.json` with almost the exact same content, right? is that intended?

static/cypress/integration/task_view.spec.ts
```diff
@@ -38,7 +47,8 @@ describe("TaskView", () => {
       intercept_next_image: intercept_next_image_with_failure,
     });
     login()
-      .wait("@get_task")
+      .wait("@get_tasks")
```
n-koenig, line 54: No critique, just for my understanding: why don't we need the `.wait("@get_tasks")` in the three tests above?

static/openapi.json
```diff
@@ -11,17 +11,20 @@
...
-                                    "ecg-qrs-classification-physiodb"
-                                ]
+                                "schema": {
+                                    "title": "Response Get Tasks",
+                                    "type": "array",
+                                    "items": {
+                                        "$ref": "#/components/schemas/Task"
+                                    }
+                                }
                             }
```
n-koenig, line 28:
```suggestion
                            },
                            "example": [
                                  "ecg-qrs-classification-physiodb",
                                  "pressure-mat-pose-estimation-slp"
                            ]
```
No must have, just for completion

static/src/App.vue
```diff
@@ -66,12 +76,18 @@ export default defineComponent({
...
         @login="login"
       ></LoginView>
+      <TasksOverView
+        v-else-if="page == 'tasks'"
+        id="tasks-overview"
+        @annotate="annotate"
+      ></TasksOverView>
       <TaskView
-        v-else
+        v-else-if="page == 'task' && task"
```
n-koenig, line 85: They way we handle things at the moment means that everything is displayed at the same base URL, right? So the browser-back/forth-buttons are useless and (except for logout) we don't have any "control" buttons yet. 
So right now, its impossible to get back to the overview after you clicked on a task, except for reloading the entire page or logging out.

Definitely shouldn't change that in this PR but maybe something that we should think about, as we have 3 completely different things to display on the same page.

static/src/components/TaskCard.vue
```diff
@@ -0,0 +1,25 @@
...
+</script>
+
+<template>
+  <i-card>
+    <template #header>
+      {{ task.id }}
+    </template>
+    {{ task.description }}
+    <template #footer>
+      <i-button color="primary" @click="$emit('annotate', task)">
```
n-koenig: Nice-to-have: I would suggest renaming this function to something like `open_task` or `start_session/task`, as this button doesn't have anything to do with an actual annotation yet. But I'm also open to your thoughts or just keeping it like this :)

static/src/components/TaskView.vue
```diff
@@ -17,48 +17,33 @@ export default defineComponent({
...
-      if (typeof OpenAPI.TOKEN === "string") {
-        return parseJwt(OpenAPI.TOKEN)["fullname"];
-      } else {
-        return undefined;
-      }
-    },
-  },
-
   mounted() {
-    this.fetchLabels();
```
n-koenig, line 45: just for my understanding, why is this not needed anymore?

n-koenig: Ah my bad, found it :)

static/src/App.vue
jb3rndt: Yep, totally agree. I should be fixed with #127

static/openapi.json
jb3rndt: I dont know how to do that, because this file gets overriden when the api is regenerated. @bewee do you know where this example has to be added in fastapi? Additionally, isnt that an example for the case when just the ids were returned? In this pr, the whole task is returned now, but the concept remains of adding a list of whole tasks

static/cypress/fixtures/tasks.json
jb3rndt: Yeah no not really, that is not good. I looked for another way and found one, which can be applied to the other tests too. But since it required refactoring in some other places too, I put it in [another pr](https://github.com/team-folivora/annoto/pull/132), to not pollute this one

static/cypress/integration/task_view.spec.ts
jb3rndt: Good question, I could not observe any difference in the test runs, so I added the line to the 3 tests where it was missing, for consistency 👍

static/src/components/TaskCard.vue
jb3rndt: `open_task` sounds nice for me :) 👍
(I am just not sure, if the convention would suggest `open-task`, `open_task` or `openTask`)
Edit: just googled it and in the vue tutorial, one would write `emit("openTask", ...)` and when using the `TaskCard`: `@open-task`. So it does the conversion automatically.

static/cypress/integration/task_view.spec.ts
bewee: Probably we there already wait for something different that happens afterwards, so that get_tasks is waited for implicitly. But honestly idk, adding it is not gonna break anything so feel free to do it :)

static/openapi.json
bewee: Yes, definitely don't change this file by hand. Don't know how to add it to fastapi right now, let's have a look at it together tomorrow.

n-koenig (approved):

static/src/App.vue
```diff
@@ -66,12 +76,18 @@ export default defineComponent({
...
         @login="login"
       ></LoginView>
+      <TasksOverView
+        v-else-if="page == 'tasks'"
+        id="tasks-overview"
+        @open-task="openTask"
+      ></TasksOverView>
       <TaskView
-        v-else
+        v-else-if="page == 'task' && task"
```
jdsauer, line 85: 'task' && task", this confuses me somehow. maybe this is a leftover or does it require the check for a string and also an object?

jdsauer (approved):

static/src/App.vue
jb3rndt: Good question, I'll take that to #127 for discussion

PR #120, opened 2022-05-16
Fix development config

bewee:
Which lead to the well-known error
```
SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 139682474104640 and this is thread id 139682407204608.           
```

### DIFF
--- a/api/.env.development
+++ b/api/.env.development
@@ -1,0 +1,1 @@
+engine_connect_args = {"check_same_thread": false}
--- a/api/.env.dev
+++ b/api/.env.dev
@@ -1,1 +1,0 @@
-engine_connect_args = {"check_same_thread": False}


api/.env.development
```diff
@@ -0,0 +1 @@
+engine_connect_args = {"check_same_thread": false}
```
jb3rndt, line 1:
```suggestion
engine_connect_args = {"check_same_thread": False}
```
Dont know if we want python-style here, but if no just ignore this comment and merge if you want

jb3rndt (approved):

api/.env.development
bewee: With the big F, it's no longer a valid json and the whole app will just crash on startup. So let's leave it^^

PR #121, opened 2022-05-16
Fix manage.py db-create not applying migrations

bewee:
Which lead to the test database entries not being loaded on docker container startup


jb3rndt (approved):
👍

PR #122, opened 2022-05-16
Remove files

jb3rndt:

- Remove the sample folder and some sloth images

I think we had there files for testing around with debug routes and the ImageDisplay. But since we have the ecg task folder, these files are not needed anymore


jb3rndt:
The only part that is missing, is a subsubfolder... So this cannot be fulfilled with the task folder that exists already. Going to close it again for now.

PR #127, opened 2022-05-17
Implement Vue router for routing

### LINKED ISSUE #123: Register helpers to vue instance
This way, we can call `this.$logout` from within any component and without imports

### LINKED ISSUE #124: Helper function for checking login status

### LINKED ISSUE #126: Look at vue state management for global state variables like fullname and jwt token

jb3rndt:

- Introduces Vue router

### Definition of Done

- [x] Code is documented
- [x] User Manual is updated


static/src/components/TaskView.vue
```diff
@@ -50,13 +56,13 @@ export default defineComponent({
...
 <template>
-  <i-layout-content>
+  <div id="task-view">
     <ProofOfCondition
       v-model:isAttentive="isAttentive"
       v-model:isTrained="isTrained"
     />
     <UserInformationLabel :fullname="fullname() ?? 'Unknown user'" />
-    <div v-if="imageId">
+    <div v-if="imageId && task">
```
jb3rndt: @jdsauer mentioned [here](https://github.com/team-folivora/annoto/pull/119#discussion_r880149038) that this is somewhat weird, and I think we dont need the check for the `task` object because it gets checked in the nextImage method. But maybe we should implement an earlier check to ensure the `TaskView` is not opened with a non-existing taskId?

static/src/utils/store.ts
```diff
@@ -0,0 +1,31 @@
...
+  set jwt(newJWT) {
+    this._jwt = newJWT;
+    this.isLoggedIn = !!newJWT;
+    OpenAPI.TOKEN = newJWT;
+    if (newJWT) {
+      this.$cookies?.set("jwt", newJWT);
+    } else {
+      this.$cookies?.remove("jwt");
+    }
+  },
```
n-koenig: So if I call the `set jwt` without any parameter, that deletes the existing jwt, right?

static/src/App.vue
```diff
@@ -1,57 +1,40 @@
...
-    LoginView,
-    TaskView,
-  },
   data() {
     return {
-      page: "login",
       task: undefined as Task | undefined,
+      ready: false,
+      store,
       loggedIn: false,
```
n-koenig: I think this is no longer used since it is now a property of the store, right?

static/src/App.vue
```diff
@@ -1,57 +1,40 @@
...
+        store.jwt = undefined;
       }
-    },
-
-    openTask(task: Task) {
-      this.task = task;
-      this.page = "task";
-    },
+    } else {
+      store.jwt = undefined;
```
n-koenig: Could/should we also call the `logout()`-method here?

static/src/components/LoginView.vue
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
 import { defineComponent } from "vue";
 import { LoginService as API } from "@/api/services/LoginService";
+import { store } from "@/utils/store";
 
 export default defineComponent({
   emits: ["login"],
```
n-koenig: I think this is also no longer needed, right?

static/src/main.ts
```diff
@@ -24,4 +30,36 @@ app.use(Toaster);
 
 app.use(VueCookies, { expire: "1d" });
 
+const routes = [
+  { path: "/", component: RedirectView },
+  {
+    path: "/login",
+    component: LoginView,
+    beforeEnter: () => {
+      return store.isLoggedIn ? "/tasks" : true;
```
n-koenig: Do I understand correctly that this directly redirects me to the task-overview if I'm already logged in and if not just continues to send me to the login page?

static/src/components/TaskView.vue
```diff
@@ -66,15 +72,15 @@ export default defineComponent({
           :label="label"
           :is-attentive="isAttentive"
           :is-trained="isTrained"
-          :competency="competency"
-          :task-id="task.id"
-          :image-id="imageId"
+          :competency="'Prof. Dr. Med.'"
```
n-koenig, line 106: Any particular reason why that is now again hardcoded inside the component and not coming from outside? Or just temporary, as we have not dealt with the competency seriously yet?

n-koenig (approved):

static/src/utils/store.ts
n-koenig: (What I actually wanted to say: is it common to do it that way or is it maybe better to define a `delete_jwt` method()?)

static/src/App.vue
n-koenig: This refers only to the `loggedIn` parameter

static/src/components/LoginView.vue
n-koenig:
```suggestion
```

static/src/App.vue
n-koenig:
```suggestion
```

static/src/components/TaskView.vue
n-koenig: Will be solved with #128

jdsauer (approved):
I didn't notice anything unusual, so this gets an approve

static/src/components/RedirectView.vue
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
+import { defineComponent } from "vue";
+
+export default defineComponent({
+  async mounted() {
+    const jwt = this.$cookies.get("jwt");
```
n-koenig: maybe we could also just use the `store.isLoggedIn` parameter here

n-koenig (changes requested):

static/src/utils/store.ts
bewee: > So if I call the `set jwt` without any parameter, that deletes the existing jwt, right?

Not when we call it without parameters, but whenever we call it with `undefined` as the value for newJWT.
To my mind it's fine to do it that way, and I would keep the functionality of the setter as it is. But I do see the benefit concerning understandability of providing an additional `removeJwt()` function which just calls `setJwt(undefined)` and think it is a good idea to add something like this

static/src/components/RedirectView.vue
bewee: That sounds reasonable

static/src/utils/store.ts
bewee: Actually this is how it's already done :sweat_smile:

static/src/App.vue
```diff
@@ -22,17 +22,17 @@ export default defineComponent({
       try {
         await PingService.ping();
       } catch (_) {
-        store.jwt = undefined;
```
bewee: I'm not completely sure, but maybe we could just have written `delete store.jwt` instead of `store.jwt = undefined` here, rendering the additional function unrequired?

static/src/components/TaskView.vue
```diff
@@ -16,26 +17,31 @@ export default defineComponent({
...
+      const id = this.$route.params["taskId"];
+      if (typeof id !== "string") return;
+      this.task = await API.getTask(id);
+      await this.nextImage();
+    },
+
     async nextImage() {
+      if (!this.task) return;
       this.imageId = await API.getNextImage(this.task.id).catch((e) => {
         console.log(e);
```
n-koenig:
```suggestion
```

n-koenig (approved):

static/src/components/TaskView.vue
jb3rndt: Okay we ended up doing some fancy error handling and stuff, so this should be safer now. (Dont know if it is better undestandable though)

static/src/App.vue
jb3rndt: Doesnt work sadly :(

n-koenig (approved):

PR #129, opened 2022-05-18
Feat/helpers access

### LINKED ISSUE #123: Register helpers to vue instance

jb3rndt:

- Converts store to a plugin and registers it on the app


static/src/main.ts
```diff
@@ -24,4 +31,38 @@ app.use(Toaster);
...
+    component: LoginView,
+    beforeEnter: () => {
+      return store.isLoggedIn ? "/tasks" : true;
+    },
+  },
+  {
+    path: "/tasks",
+    component: TasksOverView,
+    beforeEnter: () => {
+      return store.isLoggedIn ? true : "/login";
```
jb3rndt: We could replace `store` with `(app.config.globalProperties.$store as StoreApi)`

static/src/App.vue
```diff
@@ -12,27 +11,26 @@ export default defineComponent({
...
       task: undefined as Task | undefined,
       ready: false,
-      store,
     };
   },
 
   async mounted() {
-    store.initialize(this.$cookies);
-    if (store.jwt) {
+    this.$store.initialize(this.$cookies);
```
bewee: Can we simply remove this line?

static/src/main.ts
```diff
@@ -15,7 +15,8 @@ import LoginView from "./components/LoginView.vue";
 import TasksOverView from "./components/TasksOverView.vue";
 import TaskView from "./components/TaskView.vue";
 import RedirectView from "./components/RedirectView.vue";
-import { store } from "@/utils/store";
+import Store from "@/utils/store";
+import {store} from "@/utils/store";
```
bewee: ?

static/src/utils/store.ts
```diff
@@ -1,20 +1,21 @@
...
 
   initialize(cookies: VueCookies) {
     this.$cookies = cookies;
     this.jwt = cookies.get("jwt");
   },
 
-  get jwt() {
+  get jwt(): string | undefined {
     return this._jwt;
   },
```
bewee: Check again whether all this additional typing information is really necessary

bewee (approved):
Nice nice!

static/src/main.ts
jb3rndt: We'll keep it as is

PR #132, opened 2022-05-23
Improve e2e test fixture objects

jb3rndt:

- Use objects for fixtures and not the json files


bewee:
"Fix/e2e-tests": Have they not worked beforehand? 🙈 😅

bewee (approved):
However it comes that this approach suddenly works - I'm quite happy with it :)

static/cypress/integration/utils.ts
```diff
@@ -24,63 +25,54 @@ function intercept_with_spy(
...
+  cy.intercept("GET", `${Cypress.env("API_URL")}/tasks/${task.id}`, task).as(
+    "get_task"
+  );
 }
 
 export function intercept_get_tasks() {
-  cy.intercept("GET", `${Cypress.env("API_URL")}/tasks`, {
-    fixture: "tasks.json",
-  }).as("get_tasks");
+  cy.intercept("GET", `${Cypress.env("API_URL")}/tasks`, [task]).as(
```
bewee: We could consider extracting `[task]` into an object called `allTasks` or similar within `fixture_objects.ts`. But as it is the only appearance of it right now, I'm also completely fine leaving it as it is

jb3rndt:
> "Fix/e2e-tests": Have they not worked beforehand? 🙈 😅

Haha, the new title is way better :D

static/cypress/integration/utils.ts
n-koenig: Even if it is just a minor change I think it would make sense to do that and maybe to also include an object  for the "sloth.jpg", similar to the way it was earlier stored as 'nextImage.txt'. This way we don't hardcode any names into the tests but instead have all names only in the `fixture_objects.ts` (if I understood the concept correctly).
I could probably also manage to do that myself tomorrow morning :)

n-koenig (changes requested):

n-koenig (approved):

static/cypress/integration/utils.ts
jb3rndt: Good idea 👍
I quickly did that also while implementing bewee's suggestion

PR #133, opened 2022-05-23
Fix/test utils camel case

jb3rndt:

- The e2e test util functions had their names in pythonic underscore case (or what that is called) and for javascript I think camelCase is the convention (the cypress api also uses camelCase), so I thought it would fit to change that

#119 has to be merged first (sorry for not squashing)


bewee:
It's way better like this! 👍

bewee (approved):

jdsauer (approved):
makes the code much more pleasant to read and also adheres to the convention, good job!

PR #135, opened 2022-05-24
Only upload e2e videos on failure

jb3rndt:
I noticed that we might only need to upload the videos from the e2e tests when the tests fail, in other cases noone will be interested in them. But there might also be reasons to always upload them. This is just a suggestion, we dont need to do it


n-koenig (approved):

bewee (approved):
Sounds good to me too! :+1:

